PR #2, opened 2015-07-22
Require authentication for GitHub notifications

andistuder:


andistuder:
@gcapizzi please review..

spec/requests/github_notifications_controller_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      it 'updates the corresponsing RepositoryLocation' do
+        expect(RepositoryLocation).to receive(:update_from_github_notification).with(payload)
+
+        post "/github_notifications?token=#{token}", payload
+
+        expect(response).to be_ok
+      end
+
+      it 'does not create authorised session' do
+        expect(RepositoryLocation).to receive(:update_from_github_notification).with(payload)
```
gcapizzi: Let's make sure this call happens only once

andistuder: Thanks for feedback. I tried this: the current expectation actually does assert that it happens exactly once.

PR #6, opened 2015-08-07
add indexes for snapshop tables

andistuder:
Signed-off-by: Theo Cushion <tcushion@pivotal.io>


esbel:
:+1:

dideler:
Andi and I experimented with some queries using EXPLAIN in psql to verify that these indexes make a significant difference, and they do. In some cases there has to be a lot of records before the index is used by Postgres.

LGTM :+1:

PR #11, opened 2015-08-19
Releases performance

andistuder:
## What?
Improve Releases page performance

**Jira ticket**: https://jira.fundingcircle.com/secure/RapidBoard.jspa?rapidView=22&projectKey=ENG&view=planning&selectedIssue=ENG-210

## How?
By retrieving feature status from snapshots instead of processing events on the fly

## Note on deployment
This changes the way snapshots are captured. This requires the clearing and reprocessing of all old snapshots. Run `rake db:clear_snapshots` following deployment.


app/models/projections/releases_projection.rb
```diff
@@ -10,108 +10,65 @@ class ReleasesProjection
...
     def initialize(per_page:, git_repository:, app_name:)
       @per_page = per_page
-      @tickets_projection = Projections::ReleasesTicketsProjection.new
       @git_repository = git_repository
-      @feature_reviews = {}
-      @tickets_hash = {}
-      @production_deploys = {}
+      @app_name = app_name
+
+      @deploy_repository = Repositories::DeployRepository.new
```
iainbeeston: Can we pass this in as a named parameter? (with a default value of `Repositories::DeployRepository.new`)

It makes no practical difference, but it feels more consistent with how `git_repository` is initialized

iainbeeston: It might even be better not to set the default, but have the releases controller pass the `deploy_repository` like it does the `git_repository`

dideler: +1

app/decorators/feature_review_with_dependent_versions.rb
```diff
@@ -0,0 +1,16 @@
+class FeatureReviewWithDependentVersions < SimpleDelegator
+  def initialize(feature_review)
+    super(feature_review)
+    @feature_review = feature_review
```
iainbeeston: Do we still need to store a reference to `@feature_review` here? (and in the other decorators too)

With `SimpleDelegator` you can use `__getobj__()` to access the original object

iainbeeston: (or just call `super.something()`)

app/models/queries/release_query.rb
```diff
@@ -0,0 +1,48 @@
+require 'repositories/feature_review_repository'
+
+module Queries
+  class ReleaseQuery
+    def initialize(release:, git_repository:, at: Time.now)
+      @release = release
+      @time = at
+
+      @feature_review_repository = Repositories::FeatureReviewRepository.new
```
iainbeeston: Could we pass this in as a named parameter? (just like with the releases projection)

app/models/queries/release_query.rb
```diff
@@ -0,0 +1,48 @@
...
+      @git_repository = git_repository
+    end
+
+    def feature_reviews
+      feature_reviews_with_dependent_versions
+        .select { |fr| fr.dependent_versions(git_repository).include?(release.version) }
+        .map { |fr| FeatureReviewWithStatuses.new(fr, at: time) }
```
iainbeeston: Could this be done inside `feature_reviews_with_dependent_versions`? To me it feels like there should be a method that says "give me the feature reviews" and it returns them completely initialised (and decorated) and ready to use.

iainbeeston: Or could it be done in a new helper method. My main worry is that it feels odd that we decorate feature reviews with version info in `feature_reviews_with_dependent_versions` but with status info here in `feature_reviews` itself (feels like both should be handled in the same way)

app/models/repositories/deploy_repository.rb
```diff
@@ -13,12 +13,22 @@ def deploys_for(apps: nil, server:, at: nil)
       deploys(apps, server, at)
     end
 
+    def deploys_for_versions(versions, environment:)
+      query = store.select('DISTINCT ON (version) *')
```
iainbeeston: Could this be rewritten without using arel? `FeatureReviewRepository.feature_reviews_for()` is fairly similar (I think?) For example, `store.where('versions && ARRAY[?]::varchar[]', versions).where(environment: environment).order(version: :asc, id: :desc)`

iainbeeston: Alternatively you can just rewrite the `IN` clause like this: `.where('version IN (?)', versions)`

app/models/repositories/feature_review_repository.rb
```diff
@@ -9,14 +9,20 @@ def initialize(store = Snapshots::FeatureReview)
...
-        Factories::FeatureReviewFactory.new.create(
-          url: latest_snapshot.url,
-          versions: latest_snapshot.versions,
-        )
-      }
+    def feature_reviews_for(versions:, at: nil)
+      query = at ? store.arel_table['event_created_at'].lteq(at) : nil
+
+      store
+        .where(query)
```
iainbeeston: Can we rewrite this as:

```
query = at ? store.where("event_created_at <= ?", at) : store
query.where('versions && ARRAY[?]::varchar[]', versions)
```

iainbeeston: (I'm very wary of directly referencing arel in rails code!)

app/models/repositories/feature_review_repository.rb
```diff
@@ -9,14 +9,20 @@ def initialize(store = Snapshots::FeatureReview)
...
+        .where('versions && ARRAY[?]::varchar[]', versions)
+        .group_by(&:url)
+        .map { |_, snapshots|
+          most_recent_snapshot = snapshots.max { |s1, s2| s1.event_created_at <=> s2.event_created_at }
```
iainbeeston: You can write this as `snapshots.max(&:event_created_at)` - which then calls `event_created_at` on each snapshot and compares the result

app/models/repositories/deploy_repository.rb
dideler: We're using Arel for database abstraction, so there's no need to speak a specific dialect of SQL. Is there any reason why you'd rather not use Arel?

app/models/repositories/feature_review_repository.rb
dideler: See [earlier comment](https://github.com/FundingCircle/shipment_tracker/pull/11/files#r38746880)

app/models/repositories/feature_review_repository.rb
dideler: Will the shorthand know to return `a <=> b`?

```ruby
irb(main):011:0> Snapshots::Ticket.all.max { |s1, s2| s1.event_created_at <=> s2.event_created_at }
  Snapshots::Ticket Load (2.3ms)  SELECT "tickets".* FROM "tickets"
=> #<Snapshots::Ticket id: 1168, key: "ENG-49", summary: "Test", status: "In Progress", urls: [], event_created_at: "2015-09-02 10:43:04", versions: []>
irb(main):012:0> Snapshots::Ticket.all.max(&:event_created_at)
  Snapshots::Ticket Load (3.0ms)  SELECT "tickets".* FROM "tickets"
ArgumentError: wrong number of arguments (1 for 0)
```

dideler:
Thanks for reviewing @iainbeeston. The comments I haven't replied to are welcome to be made.

app/models/repositories/feature_review_repository.rb
iainbeeston: Sorry, that should have been `max_by(&:event_created_at)` (not `max(&:even_created_at)`). [`max_by`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-max_by) calls the method for each element in the enumerable, then compares the results (effectively with `<=>`)

iainbeeston: You need to use `max_by` rather than `max`.

It does effectively use the spaceship operator (the [Comparable module](http://ruby-doc.org/core-2.2.2/Comparable.html) uses `<=>` as the basis for all other comparison methods, e.g. `<` or `>`)

app/models/repositories/deploy_repository.rb
iainbeeston: @dideler Generally rails uses active record as the database abstraction, which happens to use arel underneath, but almost all of it is wrapped by active record classes and methods. I'd argue that using arel directly (circumventing active record) makes the code less readable and less maintainable, for example  `.where(store.arel_table['event_created_at'].lteq(at))` is more obfuscated than `.where("event_created_at <= ?", at)`.

However, it seems that this is an established practice in the codebase. I'd still rather we tried to move away from it going forwards, but I'll leave it up to @Neilos if he wants to change what is already working here.

app/models/queries/release_query.rb
Neilos: I can see what you're saying: from a testing point of view it would be good to do. However, the nice thing about these query objects (ReleaseQuery and the FeatureReviewQuery) is that they hide the repository dependencies from the decorator objects that use them and thus they simplify the decorators. I'm going to leave this as it is, if that's ok with you.

app/models/repositories/deploy_repository.rb
Neilos: Thanks @iainbeeston. I get what you're saying, but this project already uses arel tables elsewhere so I'll stick with the convention established for this project.

app/decorators/feature_review_with_dependent_versions.rb
Neilos: Yep. We're not using it anyway.

Neilos:
I believe this branch is ready for merge, and, as agreed, by Chris (product owner), can be deployed to production.

andistuder:
Great. LGTM. Feel free to merge once ready to deploy. Pls remember to reset the snapshots ofter deploy.

dideler:
Thanks everyone, going to deploy this now.

PR #13, opened 2015-08-28
Pull request status

iainbeeston:
WHAT?
---

This enables support for github pull request statuses, so that the feature review status in shipment tracker is shown in the status of a pull request.

HOW?
----

1. When a pull request is created or pushed to, github sends an event to shipment tracker
2. We check the status of a feature request, and POST it to github's status API endpoint

We also update the status in github whenever the status of a jira ticket is changed.

TESTING
------

To test this you need to set up a feature review in https://shipment-tracker.herokuapp.com for an app, which has it's repository hooks pointing at the staging shipment tracker (not the production shipment tracker). As it is, events will get sent to the production shipment tracker, not the staging one (which knows how to set statues).


iainbeeston:
One thing that this pull request doesn't do is set the github status for new feature requests. If a user creates a pull request before their feature request, then it won't have a status in github (unless they then push more commits to it)

Neilos:
Should probably update the `.env.development.example` file with the `GITHUB_REPO_STATUS_ACCESS_TOKEN`.

Gemfile.lock
```diff
@@ -143,6 +145,8 @@ GEM
       multi_json (~> 1.3)
       multi_xml (~> 0.5)
       rack (~> 1.2)
+    octokit (4.0.1)
```
andistuder: Version 4.1 has been released. Please upgrade.

andistuder: Version 4.1 has been released. Please upgrade.

iainbeeston:
If there's any more feedback on this please let me know. I've tweaked it some more

Procfile
```diff
@@ -1,2 +1,3 @@
 web: bundle exec rails server --port $PORT
 worker: bundle exec rake jobs:update_events_loop
+background: rake jobs:work
```
andistuder: maybe call the rake job with `bundle exec`?

README.md
```diff
@@ -96,6 +96,16 @@ bundle exec rake jobs:update_git
 
 *Warning:* This recurring task should run on **every** server that your application is running on.
 
+### Enable GitHub Webhooks
```
andistuder: :trophy: for docs

andistuder:
LGTM

config/application.rb
```diff
@@ -31,10 +31,13 @@ class Application < Rails::Application
 
     config.action_dispatch.perform_deep_munge = false
 
+    routes.default_url_options = { host: ENV['HOST_NAME'] || "localhost:#{ENV['PORT']}" }
+
     config.ssh_private_key = ENV['SSH_PRIVATE_KEY']
     config.ssh_public_key = ENV['SSH_PUBLIC_KEY']
     config.ssh_user = ENV['SSH_USER']
     config.approved_statuses = ENV.fetch('APPROVED_STATUSES', 'Done,Ready for Deployment').split(/\s*,\s*/)
+    config.github_access_token = ENV['GITHUB_REPO_STATUS_ACCESS_TOKEN']
```
andistuder: :no_entry_sign: This may not work well for the dev environment, where we load the env var `GITHUB_REPO_STATUS_ACCESS_TOKEN` via .env.development file. This is being loaded in the development environment ('development.rb'), which is evaluated after application.rb

Consider setting the `github_access_token` in a dedicated initializer, following this example here: https://github.com/FundingCircle/shipment_tracker/blob/master/config/initializers/omniauth.rb#L3

PR #14, opened 2015-09-08
Uri error

Neilos:
Fixes bug whereby the uri was not being validated upon entry but was later causing an invalid uri error to be thrown when a feature review was created.


app/models/git_repository_location.rb
```diff
@@ -1,4 +1,14 @@
 class GitRepositoryLocation < ActiveRecord::Base
+  validate :must_have_valid_uri
+
+  def must_have_valid_uri
+    URI.parse(uri)
```
iainbeeston: @Neilos is this the correct way to validate this? I thought we said that we should be parsing uris using addressable? In which case should we should validate that it parses with Addressable here? (Also, should we update code elsewhere to use Addressable::URI instead of URI?)

Neilos: That is what we discussed, but I actually I found that later functionality is dependent on it being a valid URI (i.e. not just a valid Addressable::URI).

iainbeeston: Fair enough.

There is one more thing though: I just remembered that URI provides a regexp for this kind of thing, I should have mentioned that before. Rather than trying to catch an error, could we please check against `URI::regexp` instead? (eg. `if !uri.match(URI::regexp) errors.add(...)`). We should avoid throwing and catching errors if possible - it's slow and in this case we're effectively using an error to act as a conditional statement.

iainbeeston:
@neilos Can we update this and get it merged? (Sorry, I realise you're busy pairing but it's so close to beig merged!)

app/models/git_repository_location.rb
Neilos: @iainbeeston 

URIs of the form `git@github.com:some/other-repo.git` should be invalid, but matching against `URI::regexp` does not indicate it's invalidness.
```
'git@github.com:some/other-repo.git'.match(URI::regexp)
 #=> #<MatchData "github.com:some/other-repo.git" ...>
```

Only by parsing it do we reveal the problem with the URI.
```
URI.parse('git@github.com:some/other-repo.git') 
 # => URI::InvalidURIError: bad URI(is not URI?)
```

iainbeeston: Ok, it's good that you checked. I know there are multiple URI regexps but I
think that if we rely on one it should be the main one (URI::regexp).

Than I think just merge if you're ready.

PR #15, opened 2015-09-10
[HACK WEEK] Create an endpoint to return events of a given type and also push deploy events

shabankarumba:


app/controllers/events_controller.rb
```diff
@@ -12,6 +12,39 @@ def create
...
+  def index
+    event_type = params[:event_type]
+    type = {
+      'deploy' => 'Events::DeployEvent',
+      'jenkins' => 'Events::JenkinsEvent',
+      'circle_ci' => 'Events::CircleCiEvent',
+      'manual_test' => 'Events::ManualTest',
+      'uat' => 'Events::UatEvent',
+      'jira' => 'Events::Jira',
+
```
MiroslavCsonka: Unnecessary empty line

app/controllers/events_controller.rb
```diff
@@ -12,6 +12,39 @@ def create
...
+    }[event_type]
```
MiroslavCsonka: Maybe we could use .fetch() with some default? Or it isn't problem when the `type` is `nil`?

app/services/push_events.rb
```diff
@@ -0,0 +1,46 @@
...
+      unpushed_events.update_all(pushed: true) if response.code == '201'
+    end
+  end
+
+  private
+
+  attr_reader :event_type
+
+  def body
+    return {} unless events.count > 0
```
MiroslavCsonka: Could we do `return {} if events.count.zero?`

app/services/push_events.rb
```diff
@@ -0,0 +1,46 @@
+class PushEvents
+  def self.call
+    new.call
+  end
+
+  def call
+    unless body.empty?
```
MiroslavCsonka: Maybe we could do `if body.any?`. I think that it is way clearer, what do you think?

app/services/push_events.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def body
+    return {} unless events.count > 0
+    {
+      events: events
+    }
+  end
+
+  def unpushed_events
+    @unpushed_events = Events::BaseEvent
```
MiroslavCsonka: Maybe we could memoize it?

shabankarumba: Done

app/services/push_events.rb
shabankarumba: Done

app/services/push_events.rb
shabankarumba: Done

app/controllers/events_controller.rb
shabankarumba: It still works fine

app/controllers/events_controller.rb
shabankarumba: Done

iainbeeston:
This has created a number of different errors on staging, mostly related to app/services/push_events.rb

@shabankarumba can you please take a look at [honeybadger](https://app.honeybadger.io/projects/44071/faults?resolved=f) and see if you can resolve them? (We wouldn't want to merge this until those are fixed)

app/controllers/events_controller.rb
```diff
@@ -12,6 +12,38 @@ def create
     self.response_body = 'ok'
   end
 
+  def index
+    event_type = params[:event_type]
+    type = {
+      'deploy' => 'Events::DeployEvent',
+      'jenkins' => 'Events::JenkinsEvent',
+      'circle_ci' => 'Events::CircleCiEvent',
+      'manual_test' => 'Events::ManualTest',
```
sgerrand: Should probably have `Event` appended to match the naming used elsewhere.

app/controllers/events_controller.rb
```diff
@@ -12,6 +12,38 @@ def create
...
+      'uat' => 'Events::UatEvent',
+      'jira' => 'Events::Jira',
```
sgerrand: As per the previous comment.

.rubocop.yml
```diff
@@ -39,4 +39,11 @@ Metrics/AbcSize:
 Style/MultilineBlockChain:
   Enabled: false
 
+Metrics/MethodLength:
+  Enabled: false
+
+Style/GuardClause:
+  Enabled: false
+
```
iainbeeston: @andistuder @dideler @Neilos Are you guys ok with these changes to the styleguide? (I'd rather not, furthermore I'd suggest any changes to the styleguide should be a separate pull request)

app/controllers/events_controller.rb
```diff
@@ -12,6 +12,38 @@ def create
...
+    }[event_type]
+    paginated_events = Events::BaseEvent
+                       .where(type: type)
+                       .order('created_at DESC').page(params[:page]).per_page(30)
+
+    events = paginated_events.map { |event|
```
iainbeeston: Feels like this should be defined in `BaseEvent` as `BaseEvent#as_json`? (The controller shouldn't have intimate knowledge about how to render an event)

app/controllers/events_controller.rb
```diff
@@ -12,6 +12,38 @@ def create
...
+    event_type = params[:event_type]
+    type = {
+      'deploy' => 'Events::DeployEvent',
+      'jenkins' => 'Events::JenkinsEvent',
+      'circle_ci' => 'Events::CircleCiEvent',
+      'manual_test' => 'Events::ManualTest',
+      'uat' => 'Events::UatEvent',
+      'jira' => 'Events::Jira',
+    }[event_type]
+    paginated_events = Events::BaseEvent
```
iainbeeston: It feels like this is something that should be handled by the repository objects. The `Updater` already maintains a list of repositories, and we generally load events via their repository.

Normally I'd have expected the controller to do something like:

```
repo = Updater.repository_for(event_type)
repo.all_events_for(page: params[:page], per_page: 30)
```

This would also mean that all repositories would have have an `events_for` method (I'd suggest having the `Updater` iterate through all repositories and calling `events_for` on each, in a similar way to how the `apply` method works, returning `[]` if the event type is not right for that repo)

app/models/events/base_event.rb
```diff
@@ -3,6 +3,7 @@
 module Events
   class BaseEvent < ActiveRecord::Base
     self.table_name = 'events'
+    self.per_page = 30
```
iainbeeston: This is not something the model should know or care about - it's a controller concern. I'd strongly recommend moving it out of the model

app/services/push_events.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def body
+    return {} if events.count.zero?
+    {
+      events: events,
+    }
+  end
+
+  def unpushed_events
+    @unpushed_events ||= Events::BaseEvent
```
iainbeeston: This should probably be in the `DeployRepository`

app/services/push_events.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def call
+    if body.any?
+      response = RestClient.post(
+        'http://hw-meter.meteor.com/deploy',
+        body.to_json,
+        content_type: :json,
+        accept: :json,
+      )
+      unpushed_events.update_all(pushed: true) if response.code == '201'
```
iainbeeston: This should probably be handled by the `DeployRepository`

andistuder:
@shabankarumba can we close this PR? I don't think the suggested endpoint quite what we want to serve.

shabankarumba:
Closing this was a proof  of concept

Issue #16, opened 2015-09-14
Migration fails 


andistuder:
Running migrations form an empty database fails. 

```ruby 
rake aborted!
NameError: uninitialized constant RenameRepositoriesToRepositoryLocations
```

```ruby 
rake aborted!
NameError: uninitialized constant AddRemoteHeadToRepositoryLocations
```

RenameRepositoriesToRepositoryLocations
```ruby
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

uninitialized constant Events::Base/Users/andi/Development/shipment_tracker/db/migrate/20150507103704_rename_events.rb:13:in `block in up'
```

I believe these error where introduced in the two commits where migrations where edited: 
https://github.com/FundingCircle/shipment_tracker/commit/f50b6f212a0c6d1ba47ba2cda10f02a10d5f0e6c
https://github.com/FundingCircle/shipment_tracker/commit/35d677af9514dbb8b4ee80ebad8353ae298616d6

I opened a PR with a suggested fix https://github.com/FundingCircle/shipment_tracker/pull/17

fundingcircleaudit:
solved

sgerrand:
@fundingcircleaudit: Not sure you should be logged into the web UI. :person_with_pouting_face:

andistuder:
@sgerrand my fault. I had to log in to issue a token... and forgot to switch account. AFAIK the comment above is the only action I've taken as @fundingcircleaudit

PR #17, opened 2015-09-14
fix database migrations

andistuder:
addresses issue: https://github.com/FundingCircle/shipment_tracker/issues/16

- fixes violation Rails migration naming convention
- fixes reference of Events::BaseEvent class

### DIFF
--- a/db/migrate/20150428153913_rename_repositories_to_repository_locations.rb
+++ b/db/migrate/20150428153913_rename_repositories_to_repository_locations.rb
@@ -1,4 +1,4 @@
-class RenameRepositoriesToGitRepositoryLocations < ActiveRecord::Migration
+class RenameRepositoriesToRepositoryLocations < ActiveRecord::Migration
   def change
     rename_table :repositories, :repository_locations
   end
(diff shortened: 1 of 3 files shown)


iainbeeston:
LGTM (but it's failing right now because circle can't post the results to shipment tracker, because the SSL certificate has expired)

This raises another question - should the CircleCI build fail if it can't post the results to shipment tracker?

andistuder:
> This raises another question - should the CircleCI build fail if it can't post the results to shipment tracker?

I reckon it should as we should not delivery software if we can to track the audit data. eg. I suspect notifications form Jira and on deploy will also be lost. But, feel free to raise an issue, or with PO.

andistuder:
@Neilos please review this PR.

iainbeeston:
> I reckon it should as we should not delivery software if we can to track the audit data

:+1:

Neilos:
Migrations should probably have been written in SQL rather than ruby, but the fault here was mine for not noticing the need to amend the historic migrations (and also mistakenly changing the name of the `RenameRepositoriesToRepositoryLocations` migration).
Thanks for fixing. Happy to merge.

sgerrand:
> should the CircleCI build fail if it can't post the results to shipment tracker?

@andistuder: Would be good to get this explicitly confirmed by the product owners, especially @carlpearse.

dideler:
> should the CircleCI build fail if it can't post the results to shipment tracker?
> > Would be good to get this explicitly confirmed by the product owners, especially @carlpearse.

/cc @chriswan84

PR #20, opened 2015-09-16
Releases page statuses

dideler:
https://jira.fundingcircle.com/browse/ENG-220

The ticket originally asked for changing the Releases page to
- have approved releases link to the feature review at approval time
- have unapproved releases link to the feature review at merge time
- show the time when the release was committed to master

But after our sprint planning we decided to have the current Releases page focus on the developer persona (with another page being for the auditor/assistant persona). With this the story changed to
- show current status of a release
- have approved releases link to the feature review at approval time
- have unapproved releases link to the feature review at current time (no time specified)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -2,9 +2,11 @@
 //= require jquery_ujs
 //= require bootstrap-sprockets
 //= require bootstrap-editable
+//= require bootstrap/tooltip
 
 $(function() {
   $.fn.editable.defaults.ajaxOptions = {type: "PUT"};
   $.fn.editable.defaults.mode = 'inline';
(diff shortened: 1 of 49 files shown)


dideler:
@iainbeeston @andistuder we decided to defer the rename from "path" to "request_uri" because the rename takes a bit of time and doesn't add that much value ("path" appears in many places in the codebase and in some cases does not need to be changed). @Neilos also came up with some good arguments to defer the change -- path is more commonly understood than request uri as the thing we're trying to identify (e.g. the `*_path` URL helpers from Rails, even if technically incorrect), and also path is used to locate resources (which we're doing, even if it's in the query params).

andistuder:
LGTM

PR #21, opened 2015-09-17
Whitelist query params in Feature Review URL

Neilos:
Fixes ENG-229

### DIFF
--- a/app/models/factories/feature_review_factory.rb
+++ b/app/models/factories/feature_review_factory.rb
@@ -1,9 +1,12 @@
-require 'uri'
 require 'rack/utils'
+require 'uri'
+
 require 'feature_review'
 
 module Factories
   class FeatureReviewFactory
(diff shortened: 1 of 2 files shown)


app/models/factories/feature_review_factory.rb
```diff
@@ -28,6 +31,13 @@ def create(attrs)
 
     private
 
+    def whitelisted_path(uri, query_hash)
+      white_listed_query = query_hash.select { |query_param, _|
```
iainbeeston: You can write this more succinctly as:

`query_hash.slice(QUERY_PARAM_WHITELIST)`

iainbeeston:
Apart from the style comment above this works great and looks good to merge, to me

app/models/factories/feature_review_factory.rb
Neilos: Thanks for this. We decided to use `extract!`

PR #25, opened 2015-09-21
Fix feature review approval times (and links on releases page)

Neilos:
#### Approval/unapproval events now correctly result in new feature review snapshots.

Previously the snapshots relied on there being comment.body in the jira event payload, but this meant (incorrectly) that ticket state changes did not result in a new approved_at being recorded in the feature review snapshots. This was not working because not all jira events have the same structure, and therefore comments, although included in most events, are at different node points and are sometimes an array of comments.

To fix this problem we now depend on the presence of the jira key alone (which is always present) and use this to look up a ticket and subsequently the associated feature reviews from the ticket paths attribute.

#### When snapshots are regenerated the necessary queries use the event time not the current time

Previously (when querying at the current time) the first event a feature review that was approved only recently was incorrectly registering as an approval event was causing the approved_at time to be incorrectly recorded in feature review snapshots. By ensuring all queries are as at the event created_at time we fix this bug.

#### Viewing a feature review allows for time precision differences.

The "time" query value in the Feature Review URL typically has no microseconds (i.e. 0 usec), whereas the times in records that are persisted to the DB have higher precision which includes microseconds. We now allow for this.

#### Show the approval time in feature status 
On the feature review page we now show the approval time (only when Feature Review approved).
This was thought to be a useful addition to the page, and also made it easier to test the correct setting of the approved_at time in the feature_review feature.

#### Other changes to the code

There is no longer a need to wrap a feature review in a decorator to determine whether a feature review is approved. Feature reviews now know whether they are approved or not based on whether the approved_at value is set on the feature review snapshot from which it is created.

This change required moving the updating of pull requests from the ticket repository to the feature review repository, because (now) the feature_reviews do not know they have changed state (e.g. from not_approved to approved) until the jira event is processed by the feature review repository and an approved_at is recorded in the snapshot.


app/models/repositories/ticket_repository.rb
```diff
@@ -24,6 +24,11 @@ def tickets_for(feature_review_path:, at: nil)
         .map { |t| Ticket.new(t.attributes) }
     end
 
+    def most_recent_snapshot(key = nil)
+      return store.last if key.nil?
+      store.where(key: key).last
```
andistuder: Let's add a index for key to the table, and specify order explicit `.order('id DESC')`

app/models/repositories/ticket_repository.rb
```diff
@@ -24,6 +24,11 @@ def tickets_for(feature_review_path:, at: nil)
         .map { |t| Ticket.new(t.attributes) }
     end
 
+    def most_recent_snapshot(key = nil)
+      return store.last if key.nil?
```
andistuder: not sure if store.last is needed, let's review

app/models/repositories/ticket_repository.rb
```diff
@@ -41,10 +46,6 @@ def apply(event)
       )
 
       store.create!(new_ticket)
-
-      new_ticket['paths'].each do |path|
-        update_feature_review(path)
-      end
     end
 
     def update_pull_request(app_name, version)
```
andistuder: we'll remove this method, it's now in the feature review repo

andistuder:
LGTM

PR #26, opened 2015-09-22
only show commits on master branch on release page

andistuder:


andistuder:
ping @dideler please review

app/models/queries/release_query.rb
```diff
@@ -32,17 +32,9 @@ def raw_feature_reviews
...
-    end
-
-    def descendant_versions
-      git_repository.get_descendant_commits_of_branch(release.version).map(&:id)
-    end
-
-    def parent_version
-      git_repository.branch_parent(release.version)
+      versions = [release.version]
+      versions.push(release.commit.parents[1])
```
andistuder: Dennis: maybe use `.second`

spec/models/git_repository_spec.rb
```diff
@@ -67,13 +67,19 @@
     end
   end
 
-  describe '#recent_commits' do
-    let(:git_diagram) { '-o-A-B-C' }
+  describe '#recent_commits_by_first_parent' do
```
andistuder: Add test to assert that this returns git objects with parents array.

features/releases.feature
```diff
@@ -72,13 +72,10 @@ Scenario: Viewing releases for an app
   Then I should see the "pending" releases
     | version  | subject                         | feature reviews | review statuses        | review times          | approved |
     | #merge2  | Merged `feature2` into `master` | FR_456          | Approved               | 2014-10-04 15:24:34   | yes      |
```
andistuder: Dennis: try and replicate the case of multiple feature reviews on a merge commit.

app/models/git_repository.rb
```diff
@@ -26,10 +26,11 @@ def commits_between(from, to)
     end
   end
 
-  def recent_commits(count = 50)
+  def recent_commits_by_first_parent(count = 50)
```
dideler: Do you think `recent_commits_on_master` or `recent_commits_on_main_branch` would be more clear?

app/models/git_repository.rb
```diff
@@ -128,7 +129,9 @@ def build_commit(commit)
       id: commit.oid,
       author_name: commit.author[:name],
       message: commit.message,
-      time: commit.time)
+      time: commit.time,
+      parents: commit.parents.map(&:oid),
```
dideler: Since we're storing the oids of the parents and not the commit objects, it might be more clear if it's called `parent_ids`.

app/models/queries/release_query.rb
```diff
@@ -32,17 +32,9 @@ def raw_feature_reviews
...
+      versions.compact
```
dideler: Could one-line this and it would still be readable, e.g.
```ruby
[release.version, release.commit.parent_ids.second].compact
```

spec/models/git_commit_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.describe GitCommit do
   describe '#subject_line' do
-    subject(:commit) { GitCommit.new(message: "subject\n\nlorem ipsum dolor sit amet") }
+    subject(:commit) { described_class.new(message: "subject\n\nlorem ipsum dolor sit amet") }
```
dideler: :disappointed:

spec/models/git_repository_spec.rb
```diff
@@ -67,13 +67,19 @@
...
+    let(:git_diagram) do
+      <<-'EOS'
+        B--C----E
+       /         \
+ -X---A-------D---F---G-
+        EOS
```
dideler: The commits that are not part of the expectation (X, B, C, E) could be made into `o` commits (like in the other git diagrams) to signify that they are not important.

spec/models/queries/release_query_spec.rb
```diff
@@ -103,9 +104,14 @@
...
   describe '#feature_reviews' do
     context 'when release is a merge commit on master branch (F)' do
-      let(:release) { instance_double(Release, version: sha_for_F) }
-
-      it "returns the feature reviews for itself, it's descendants and it's branch parent" do
+      let(:release) {
+        instance_double(Release,
+          version: sha_for_F,
+          commit: GitCommit.new(id: sha_for_F, parents: [sha_for_D, sha_for_E]),
+                       )
```
dideler: Looks like Rubocop autocorrected this in a not so nice way.

spec/models/queries/release_query_spec.rb
```diff
@@ -103,9 +104,14 @@
...
+      }
+
+      it "returns the feature reviews for itself, and it's branch parent" do
```
dideler: s/it's/its/ and can then single quote the string

spec/models/queries/release_query_spec.rb
```diff
@@ -154,12 +170,19 @@
     end
 
     context 'when release is feature branch commit (C)' do
-      let(:release) { instance_double(Release, version: sha_for_C) }
-
-      it "returns feature reviews for itself and it's descendants" do
+      # This case is no longer expected as we only show master commits on the releases page. Once we refactor
```
dideler: Consider prepending a `TODO:` to the comment so it's easier to find in the future (though rubocop might not like it).

app/models/git_repository.rb
andistuder: thanks, good suggestions. I go for the latter.

dideler:
Looks good so far. None of my comments are blockers except for the comments I wrote with you yesterday about adding those extra test cases.

app/models/git_repository.rb
andistuder: thanks, good point.

app/models/queries/release_query.rb
andistuder: :+1:

spec/models/git_commit_spec.rb
andistuder: I prefer it as it keeps the code dryer. I will leave it as it is, as `described_class` is already used widely across tests. But open to discuss if we should remove it from the code entirely... it's a style guide question to discuss widely.

spec/models/git_repository_spec.rb
andistuder: yes, I am aware of this. But I was keen to share repo builds. Would you prefer if I create an additional one?

spec/models/git_commit_spec.rb
dideler: My preference is to aim for [damp tests](http://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests) when possible. But let's discuss offline sometime, until then it's fine by me to leave this as is.

I think this codebase currently has specs where we use `described_class` and some where we don't, we'll just need to settle on one.

spec/models/git_repository_spec.rb
dideler: > Would you prefer if I create an additional one?

If that's okay with you. It would make the intention of the test easier to understand.

dideler:
Ready for :rocket:

PR #27, opened 2015-09-23
only post Github status updates if repository is under audit

andistuder:
At current all FundingCircle repositories get status updates on pull requests form Github. With this change we assert that the repo is under audit* before we post a status update. 

\* we assume it's under audit if listed as GitRepositoryLocation

### DIFF
--- a/app/controllers/github_notifications_controller.rb
+++ b/app/controllers/github_notifications_controller.rb
@@ -47,7 +47,12 @@ class GithubNotificationsController < ApplicationController
   end
 
   def relevant_pull_request?
-    payload.action == 'opened' || payload.action == 'synchronize'
+    (payload.action == 'opened' || payload.action == 'synchronize') && audited_repo?
+  end
+
+  def audited_repo?
+    inferred_repo_name = payload.base_repo_url.split('/').last
+    GitRepositoryLocation.uris.any? { |uri| uri.include?(inferred_repo_name) }
   end
 
   def payload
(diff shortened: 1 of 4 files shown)


andistuder:
ping @dideler @sgerrand bug fix as discussed

app/controllers/github_notifications_controller.rb
```diff
@@ -47,7 +47,12 @@ def process_pull_request
   end
 
   def relevant_pull_request?
-    payload.action == 'opened' || payload.action == 'synchronize'
+    (payload.action == 'opened' || payload.action == 'synchronize') && audited_repo?
+  end
+
+  def audited_repo?
+    inferred_repo_name = payload.base_repo_url.split('/').last
```
dideler: Not a blocker: the payload also contains the repo name which we could fetch instead of inferring it.

dideler:
LGTM :+1:

sgerrand:
@andistuder: Ironic that Shipment Tracker was failing your build. :wink: Should there have been a feature review created for this change? I'm not sure of the process being applied to this project.

app/controllers/github_notifications_controller.rb
```diff
@@ -47,7 +47,12 @@ def process_pull_request
   end
 
   def relevant_pull_request?
-    payload.action == 'opened' || payload.action == 'synchronize'
+    (payload.action == 'opened' || payload.action == 'synchronize') && audited_repo?
```
sgerrand: :memo: Alternatively
```ruby
payload.action == ('opened' || 'synchronize') && audited_repo?

app/controllers/github_notifications_controller.rb
sgerrand: Could also use `URI` methods to split out the composite elements.

andistuder:
@sgerrand Shipment Tracker was quite right here: it shows failing as the change is not approved by a product owner in Jira (yet).

PR #28, opened 2015-09-25
refactor releases project

andistuder:
preload all feature reviews to increas performance
no longer use release query to simply structure


andistuder:
@dideler please review

app/models/queries/releases_query.rb
```diff
@@ -31,8 +30,16 @@ def commits
...
+      @feature_reviews ||= feature_review_repository.feature_reviews_for_versions(associated_versions)
+    end
+
     def versions
-      @versions ||= commits.map(&:id)
+      commits.map(&:id)
+    end
+
+    def associated_versions
+      commits.map(&:associated_ids).flatten.uniq
```
dideler: Is the `uniq` necessary?

spec/models/git_commit_spec.rb
```diff
@@ -2,10 +2,44 @@
...
+    end
+  end
+
+  describe '#associated_ids' do
+    subject(:associated_ids) { described_class.new(id: id, parent_ids: parent_ids).associated_ids }
+    let(:parent_ids) { [first_parent_id, second_parent_id].compact }
+    let(:first_parent_id) { '789' }
+    let(:id) { '123' }
+
+    context 'when commit is a merge commit, resp. has a second parent' do
```
dideler: Does "resp." mean respectively? I suggest either writing it full out or as "(i.e. has a second parent)"
Same for L28.

spec/models/release_spec.rb
```diff
@@ -1,12 +1,54 @@
...
+    it 'returns true if any of its feature reviews are approved' do
+      allow(feature_review1).to receive(:approved?).and_return(true)
+      allow(feature_review2).to receive(:approved?).and_return(false)
+      expect(release.approved?).to be true
+    end
+
+    it 'returns false if none of its feature reviews are approved' do
+      allow(feature_review1).to receive(:approved?).and_return(false)
+      allow(feature_review2).to receive(:approved?).and_return(false)
+      expect(release.approved?).to eq(false)
```
dideler: `to be false` to be consistent with the example above

spec/models/release_spec.rb
```diff
@@ -1,12 +1,54 @@
...
+    end
+  end
+
+  describe '#approval_status' do
+    context 'when release has NO feature review(s)' do
+      it 'returns blank when the release has no features' do
```
dideler: Just `returns blank` should be good enough (otherwise it's redundant with the context).

spec/models/release_spec.rb
```diff
@@ -1,12 +1,54 @@
...
+        expect(release.approval_status).to be_nil
+      end
+    end
+
+    context 'when release has feature review(s)' do
+      subject(:release) { Release.new(feature_reviews: feature_reviews) }
+
+      it 'returns "approved" when the release is :approved' do
```
dideler: The symbol and string should be swapped => "returns :approved when the release is approved".
Same for L48 => "returns :not_approved when the release is not approved"

spec/models/release_spec.rb
```diff
@@ -1,12 +1,54 @@
...
+        allow(release).to receive(:approved?).and_return(true)
+        expect(release.approval_status).to eq(:approved)
+      end
+
+      it 'returns "unapproved" when the release is NOT :approved?' do
+        allow(release).to receive(:approved?).and_return(false)
```
dideler: Not sure how I feel about stubbing the object under test. Might be nicer to setup the test by stubbing the `ReleaseQuery` to say there is an approved feature review.

dideler:
LGTM, just some minor comments. Nice work!

app/models/queries/releases_query.rb
andistuder: It's very unlikely to have duplicates. With the current implementation of `commit.associated_ids`, the only situation I could think of is when the same branch is merged into the main branch twice. 

But in general, it gives a greater degree of abstraction, to call `uniq` here. ie. we don't really  have to care on what business rules are applied in  `commit.associated_ids`.

spec/models/git_commit_spec.rb
andistuder: good suggestion, will do.  thnx ;)

spec/models/release_spec.rb
andistuder: :bow:

spec/models/release_spec.rb
andistuder: Less is more, thanks.

spec/models/release_spec.rb
andistuder: good spot, thanks!

spec/models/release_spec.rb
andistuder: I prefer the stub as it separates concerns. We are testing `approval_status` here and not `approved?`. If we'd test it by stubbing in the `ReleaseQuery`, we essentially would test `approved?` twice, which would lead to brittle tests. IMO. 

But, looking at it, I think `approval_status`  is really a front-end concern, and I will move it into the  `ReleasesHelper`.

andistuder: err.... turns out, `approval_status` is not used at all. I will simply delete it.

andistuder:
ping @dideler Chris approved the task, please merge.

dideler:
Thanks @andistuder, ready for :rocket:

PR #35, opened 2015-09-30
Add Deployed as approved status

andistuder:


spec/models/ticket_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe Ticket do
+  describe '#approved?' do
```
dideler: :trophy:

PR #51, opened 2015-10-13
add help links

andistuder:
adds help links at various places in the app.

### DIFF
--- a/app/views/feature_reviews/show.html.haml
+++ b/app/views/feature_reviews/show.html.haml
@@ -1,6 +1,8 @@
 
 - if @feature_review_with_statuses.time
-  .alert.alert-warning.time= "Feature review as of #{@feature_review_with_statuses.time}"
+  .alert.alert-warning.time
+    %span.glyphicon.glyphicon-time
+    As viewed at #{@feature_review_with_statuses.time}
 
 - title 'Feature Review', help_url: 'https://github.com/FundingCircle/shipment_tracker/wiki/2.-PREPARE-A-FEATURE-FOR-REVIEW'
 %p Use the Feature Review page to ensure that the feature is in a ready state to be reviewed and signed off.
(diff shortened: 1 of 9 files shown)


app/views/feature_reviews/show.html.haml
```diff
@@ -1,10 +1,13 @@
-- title "Feature Review"
+
 - if @feature_review_with_statuses.time
-  .lead.time= "Projecting events up to #{@feature_review_with_statuses.time}"
+  .alert.alert-warning.time= "Feature review as of #{@feature_review_with_statuses.time}"
```
dideler: This text seems a bit awkward (the old text did too). Would it sound better/be more clear if we said

> Viewing Feature Review as appeared at \<DateTime\>

app/views/releases/show.html.haml
```diff
@@ -1,4 +1,12 @@
-- title "Releases for #{@app_name}"
+- title "Releases for #{@app_name}", help_url: 'https://github.com/FundingCircle/shipment_tracker/wiki/5.-MERGING-AND-DEPLOYING-A-FEATURE'
+
+%p
+  The Releases page on Shipment tracker can tell you what commits are pending deployment or have already been
+  deployed for your repository and whether they have been approved by a Product Owner. From this page you can see:
```
dideler: Could this be shortened to

> View which releases (merged commits) are pending deployment or have already been deployed, and whether any associated tickets have been approved in JIRA.

And drop the list below as it more or less repeats what was stated earlier.

chriswan84: Agreed. Lets change to what dennis said.

app/views/feature_reviews/show.html.haml
chriswan84: The text for the flash notification should be:
"[clock glyph] As viewed at [Time]"

app/views/releases/show.html.haml
chriswan84: Lets also move the page help links across to follow the page title.

PR #56, opened 2015-10-16
process links in Jira markup

andistuder:
This changes allows parsing of links posted in Jira comments with Jira markup

### DIFF
--- a/app/models/repositories/ticket_repository.rb
+++ b/app/models/repositories/ticket_repository.rb
@@ -44,15 +44,7 @@ module Repositories
 
       feature_reviews = feature_review_factory.create_from_text(event.comment)
 
-      new_ticket = last_ticket.merge(
-        'key' => event.key,
-        'summary' => event.summary,
-        'status' => event.status,
-        'paths' => merge_ticket_paths(last_ticket, feature_reviews),
-        'event_created_at' => event.created_at,
-        'versions' => merge_ticket_versions(last_ticket, feature_reviews),
-        'approved_at' => merge_approved_at(last_ticket, event),
-      )
+      new_ticket = build_ticket(last_ticket, event, feature_reviews)
 
       store.create!(new_ticket)
       update_pull_requests_for(new_ticket) if update_pull_request?(event, feature_reviews)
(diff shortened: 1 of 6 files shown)


andistuder:
@dideler @0mega pls review

andistuder:
If we want this change to be applied for existing events, then it will require re-snapshotting.

app/models/repositories/ticket_repository.rb
```diff
@@ -67,6 +59,22 @@ def update_pull_request?(event, feature_reviews)
...
+        'status' => event.status,
+        'paths' => merge_ticket_paths(last_ticket, feature_reviews),
+        'event_created_at' => event.created_at,
+        'versions' => merge_ticket_versions(last_ticket, feature_reviews),
+        'approved_at' => merge_approved_at(last_ticket, event),
+      )
+    end
+
+    def normalise_jira_markup(string)
+      string.gsub(/\[(.*)\|([\S]*)\]/, ' \1 \2 ')
```
dideler: *Not a blocker:* A global regex search with captures won't be very efficient, considering we have to perform it for every new comment and then process the same comment again inside `#create_from_text`. Alternatively we could do some efficient string processing such as a `chomp(']')` on the URI inside `FeatureReviewFactory` since the issue is that any URIs found with JIRA markup will contain a trailing `]`.

0mega: +1 for @dideler, it looks like it will be faster since you already have the url_string and just need to match one char in that string to remove it.

andistuder: thanks for feedback here, however, I am a bit cautious of doing chomp, `]` is a valid character in URI. There might be a valid link which has a `]` at the end. 

Plus, I prefer to keep the knowledge on Jira Markup in this class, as it's related to the event (resp. Jira payload), rather then the Factory.

dideler: > There might be a valid link with a `]` at the end.

In this case we control what the URL looks like and at the moment it's not a concern as no ST links contain a `]` at the end.

>  keep the knowledge on Jira Markup in this class, as it's related to the event (resp. Jira payload), rather than the Factory.

I think it depends on how you look at it. If we think of it as cleaning up any JIRA links then it makes sense to do that in the `TicketRepository` which knows all about JIRA events, but if we think of it as a URI extraction or URL validation problem it goes well in the `FeatureReviewFactory`.

The current solution is definitely easier to understand than my proposed solution (which seems a bit odd in the Factory without much context). I'm just worried that we make a part of the code which sees a lot of activity noticeably slower as the number of JIRA comments we receive increases. That said, this isn't a blocker because we probably won't notice a difference right away and we can deal with performance when it becomes more of an issue.

andistuder:
@0mega @dideler Please find the PR updated, with your suggestions taken on board.

PR #70, opened 2015-11-02
ST not finding a commit that does exist

0mega:
Some users reported a 500 error on ST search page.

This PR prevents the error from happening and logs some context when this unexpected situation would occur so that we are better informed if this still is an issue in future.


app/models/git_repository.rb
```diff
@@ -126,7 +126,15 @@ def commit_on_master?(commit_oid)
     return true unless parent_commit
 
     walker = get_walker(main_branch.target_id, parent_commit.oid, true)
-    walker.first.oid == commit_oid
+
+    first = walker.first
+    if first.nil?
```
LogaJ: This should have a corresponding context test entry.

0mega: @LogaJ  Thanks for the feedback, we are not sure under which circumstances this happens, that's why we postponed the test until we have more information (probably the logs we added will provide some more insight) instead of just writing a test which does not reproduce the context.

PR #75, opened 2015-11-06
ensure uat-url protocol is stripped out

andistuder:
This will give consistency when creating and editing feature reviews. 
The change also strips out the path as not relevant in the context of a uat url.


andistuder:
ping @dideler

andistuder:
ping @0mega

0mega:
:+1:

PR #83, opened 2016-01-12
Handle Heroku deploy payload

dideler:
* Map Heroku fields to standard deploy event fields
  because we cannot change parameters sent by Heroku

Signed-off-by: Molly Huerster <molly.huerster@fundingcircle.com>

### DIFF
--- a/app/models/events/deploy_event.rb
+++ b/app/models/events/deploy_event.rb
@@ -2,6 +2,8 @@ require 'events/base_event'
 
 module Events
   class DeployEvent < Events::BaseEvent
+    ENVIRONMENTS = %w(uat staging production)
+
     def app_name
       details.fetch('app_name', details.fetch('app', nil)).try(:downcase)
     end
@@ -19,10 +21,22 @@ module Events
     end
 
     def environment
-      details.fetch('environment', nil)
+      details.fetch('environment', heroku_environment)
     end
 
     private
+
+    def heroku_environment
+      if ENVIRONMENTS.include?(heroku_name)
+        heroku_name
+      else
+        nil
+      end
+    end
+
+    def heroku_name
+      app_name.try(:split, '-').try(:last)
+    end
 
     def servers
       details.fetch('servers', servers_fallback)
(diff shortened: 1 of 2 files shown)


dideler:
@andistuder

app/models/events/deploy_event.rb
```diff
@@ -11,25 +13,37 @@ def server
...
 
+    def heroku_environment
+      if ENVIRONMENTS.include?(heroku_name)
+        heroku_name
+      else
+        nil
+      end
+    end
+
+    def heroku_name
```
andistuder: not sure if I'd call this `heroku_name` - maybe `app_name_extension` ... but not a blocker..

app/models/events/deploy_event.rb
```diff
@@ -11,25 +13,37 @@ def server
...
+      app_name.try(:split, '-').try(:last)
```
andistuder: may just `return nil unless app_name` at beginning of method. Then you won't need all these `try`s

app/models/events/deploy_event.rb
```diff
@@ -11,25 +13,37 @@ def server
     end
 
     def version
-      details.fetch('version', nil)
+      details.fetch('version', details.fetch('head_long', nil))
```
andistuder: how about 
`details['version'] || details['head_long']`
instead of nested fetch.. ?

andistuder:
Looks good, only I was expecting you'd include 'locale' as discussed?

spec/models/events/deploy_event_spec.rb
```diff
@@ -1,23 +1,52 @@
...
+  subject { Events::DeployEvent.new(details: payload) }
 
   context 'when given a valid payload' do
+    context 'when the payload comes from Heroku' do
+      let(:payload) {
+        {
+          'app' => 'nameless-forest-uat',
+          'user' => 'user@example.com',
+          'url' => 'http://nameless-forest-uat.herokuapp.com',
+          'head_long' => '123abc',
```
andistuder: please add 'locale' attribute here.

andistuder: ignore above, I was miss-understood the heroku behaviour.

andistuder:
LGMT - pls self merge as and when you're ready.

andistuder:
LGTM.. just needs a FR ;)

PR #85, opened 2016-01-13
Re-enable CircleCI webhook event type

dideler:
This will enable "CircleCI (webhook)" as a *Source* option on the Tokens page.

Before we can stop using the circleci-manual event type, we need to
verify that the CircleCI webhook now works in production, and then
switch over all audited repos to use the official webhook.

Discussion about why the manual CircleCI webhooks exist: https://jira.fundingcircle.com/browse/ENG-86


andistuder:
LGTM

andistuder:
as agreed: we might roll-back on this if the CircleCi webhook messages are still being blocked by our production systems.

PR #86, opened 2016-01-14
Clarify difference between the two CircleCI event types

dideler:
There are now two ways to get build results from CircleCI.

1. As a webhook (new way)
1. As a post test command (old way)

Both options now show up in the Tokens page dropdown list for event source.

Related to https://github.com/FundingCircle/tech-playbook/pull/95


andistuder:
LGTM

PR #90, opened 2016-02-01
Handle inconsistent letter case for deploy event environments

dideler:
Ensures the environment of deploy events are correctly stored. Ideally we fix the casing at the source before the payload is sent, but this is a necessary fallback.

@FundingCircle/black-bat


andistuder:
LGTM

PR #93, opened 2016-02-09
Multi region Releases page

0mega:
Support displaying deploys for multiple Regions e.g. UK, US, CE.

When no filter is provided in query parameters, the default Region is UK, this is for backward compatibility.

Default locale for snapshots is set in environment variable DEFAULT_LOCALE given that older deploy events didn't have a locale assigned to it.

The selected locale/region is set in the cookie for better user experience.

### DIFF
--- a/app/models/events/deploy_event.rb
+++ b/app/models/events/deploy_event.rb
@@ -25,9 +25,7 @@ module Events
     end
 
     def locale
-      # default value needed for older events without locale
-      # value is 'gb' and not 'uk' to comply with 'ISO 3166-1 alpha-2'
-      details.fetch('locale', 'gb')
+      details.fetch('locale', heroku_locale).try(:downcase) || Rails.configuration.default_deploy_locale
     end
 
     private
@@ -36,9 +34,18 @@ module Events
       app_name_extension if ENVIRONMENTS.include?(app_name_extension)
     end
 
+    def heroku_locale
+      app_name_prefix if Rails.configuration.available_deploy_regions.include?(app_name_prefix)
+    end
+
     def app_name_extension
       return nil unless app_name
       app_name.split('-').last.downcase
+    end
+
+    def app_name_prefix
+      return nil unless app_name
+      app_name.split('-').first.downcase
     end
 
     def servers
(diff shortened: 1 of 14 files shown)


0mega:
@FundingCircle/black-bat :eyes:

app/models/queries/releases_query.rb
```diff
@@ -28,7 +29,8 @@ def versions
     attr_reader :app_name, :deploy_repository, :feature_review_factory, :git_repository, :ticket_repository
 
     def production_deploys
-      @production_deploys ||= deploy_repository.deploys_for_versions(versions, environment: 'production')
+      @production_deploys ||= deploy_repository
+                              .deploys_for_versions(versions, environment: 'production', region: @region)
```
andistuder: @dideler decided not to set `attr_reader` for `@region` - treating it same as `@per_page`

app/models/events/deploy_event.rb
```diff
@@ -24,17 +24,30 @@ def environment
       details.fetch('environment', heroku_environment).try(:downcase)
     end
 
+    def locale
+      details.fetch('locale', heroku_locale).try(:downcase) || Rails.configuration.default_deploy_locale
```
andistuder: @LogaJ decided to not protect against empty string. This keeps behaviour in line with other attributes (eg. `environment`) where we also allow empty string. 

Should we see a need for more validation, we can do this any time, and then simply re-run the snapshots.

config/application.rb
```diff
@@ -40,5 +40,12 @@ class Application < Rails::Application
     config.git_repository_cache_dir = Dir.tmpdir
     config.github_access_token = ENV['GITHUB_REPO_STATUS_ACCESS_TOKEN']
     config.data_maintenance_mode = ENV['DATA_MAINTENANCE'] == 'true'
+
+    # default value needed for older events without locale
+    # value is 'gb' and not 'uk' to comply with 'ISO 3166-1 alpha-2'
+    config.default_deploy_locale = ENV.fetch('DEFAULT_DEPLOY_LOCALE', 'gb')
+
+    config.default_deploy_region = ENV.fetch('DEFAULT_DEPLOY_REGION', 'gb')
+    config.available_deploy_regions = ENV.fetch('available_deploy_region', %w(gb us))
```
dideler: Is there a specific reason why `available_deploy_region` isn't all caps?

PR #94, opened 2016-02-10
set default heroku deploy locale

andistuder:
not all apps hosted on Heroku have the locale as name prefix. With this change we allowing it to be set ot a default locale with env var `DEFAULT_HEROKU_DEPLOY_LOCALE`, defaults to `us`.

### DIFF
--- a/app/models/events/deploy_event.rb
+++ b/app/models/events/deploy_event.rb
@@ -25,7 +25,7 @@ module Events
     end
 
     def locale
-      details.fetch('locale', heroku_locale).try(:downcase) || Rails.configuration.default_deploy_locale
+      details.fetch('locale', heroku_locale).try(:downcase) || default_locale
     end
 
     private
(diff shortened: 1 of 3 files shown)


andistuder:
ping @FundingCircle/black-bat :eyes:

0mega:
:+1: should we have the env variable description in README?

app/models/events/deploy_event.rb
```diff
@@ -38,6 +38,18 @@ def heroku_locale
...
+    def default_locale
+      if deployed_to_heroku?
+        Rails.configuration.default_heroku_deploy_locale
+      else
+        Rails.configuration.default_deploy_locale
+      end
+    end
+
+    def deployed_to_heroku?
+      details.fetch('url', '').split('.')[-2] == 'herokuapp'
```
dideler: Will this still work for Heroku production apps that have a custom domain?

andistuder: yes. the 'URL' is always set to the primary location. Custom domains are just CNAME entries.

spec/models/events/deploy_event_spec.rb
```diff
@@ -59,7 +59,9 @@
       end
 
       context 'when the app name does not have the environment in lowercase' do
-        let(:payload) { { 'app' => 'nameless-forest-UAT' } }
+        before do
+          payload['app'] = 'nameless-forest-UAT'
+        end
```
dideler: Since the expectations haven't changed, suggest keeping the `let`s as the other attributes aren't being tested.

andistuder: ooops, just seen this now, after the merge: I prefer if the test runs against a payload format that is referenced in the context.

dideler:
>  should we have the env variable description in README?

I don't think it's needed in the README since there are defaults, but would be good to add to `.env.development.example`.

PR #95, opened 2016-02-10
Document env vars for deploy regions and locales

dideler:
Addresses https://github.com/FundingCircle/shipment_tracker/pull/94#issuecomment-182395256

### DIFF
--- a/.env.development.example
+++ b/.env.development.example
@@ -1,11 +1,16 @@
 AUTH_LOGIN_URL=example_url
 AUTH_LOGOUT_URL=example_url
 
-# Or set these instead if you use auth0.com
+# Or set these instead if you use auth0.com for authentication.
 AUTH0_CLIENT_ID=example_auth0_client_id
 AUTH0_CLIENT_SECRET=example_auth0_client_secret
 AUTH0_CONNECTION=example_auth0_connection_strategy
 AUTH0_DOMAIN=example_auth0_domain_without_protocol
+
+# Locales are two-letter country codes as defined in ISO 3166-1 alpha-2 codes.
+DEFAULT_DEPLOY_REGION=two-letter_region_code
+DEFAULT_DEPLOY_LOCALE=two-letter_country_code
+DEFAULT_HEROKU_DEPLOY_LOCALE=two-letter_country_code
 
 GITHUB_REPO_STATUS_ACCESS_TOKEN=example_token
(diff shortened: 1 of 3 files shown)


config/application.rb
```diff
@@ -50,6 +50,6 @@ class Application < Rails::Application
     config.default_deploy_region = ENV.fetch('DEFAULT_DEPLOY_REGION', 'gb')
 
     # value is 'gb' and not 'uk' to comply with 'ISO 3166-1 alpha-2'
-    config.available_deploy_regions = ENV.fetch('available_deploy_region', %w(gb us))
+    config.available_deploy_regions = ENV.fetch('DEPLOY_REGIONS', 'gb,us').split(',')
```
andistuder: thanks, for fixing this!

andistuder: maybe rename the config attribute from `available_deploy_regions` to `deploy_regions`?

dideler: Agreed, will do.

LogaJ:
Maybe update the readme to mention copying the `.env.development.example` to `.env.development` it's an obvious thing but it's done for the `database.yml.example` file.

.env.development.example
```diff
@@ -1,12 +1,19 @@
...
-# Or set these instead if you use auth0.com
+# Or set these instead if you use auth0.com for authentication.
 AUTH0_CLIENT_ID=example_auth0_client_id
 AUTH0_CLIENT_SECRET=example_auth0_client_secret
 AUTH0_CONNECTION=example_auth0_connection_strategy
 AUTH0_DOMAIN=example_auth0_domain_without_protocol
 
+# Locales are two-letter country codes as defined in ISO 3166-1 alpha-2 codes.
+DEFAULT_DEPLOY_LOCALE=two-letter_country_code
+DEFAULT_HEROKU_DEPLOY_LOCALE=two-letter_country_code
```
0mega: Is it worth having a comment explaining what are those used for, or we assume people will just check the code?

dideler: I think they are well named, and if they need more info can grep the code.

The app will run without these set because of hardcoded fallbacks. Writing that made me realized that maybe these aren't well named. Should maybe remove "DEFAULT" since the real defaults are the hardcoded values.

0mega: :+1:

0mega:
LGTM

andistuder:
LGTM

PR #96, opened 2016-02-10
Multi region tabs for Releases page

0mega:
So that user can select the region for the releases

screen shot 2016-02-10 at 15 02 01

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gem 'rails', '~> 4.2.1'
 
 gem 'addressable', require: 'addressable/uri'
 gem 'bootstrap-sass'
-gem 'flag-icon-sass', :git => 'git@github.com:cfiorini/flag-icon-sass.git', :branch => 'master'
+gem 'flag-icon-sass'
 gem 'delayed_job_active_record'
 gem 'dotenv'
 gem 'haml-rails'
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ gem 'rails', '~> 4.2.1'
 
 gem 'addressable', require: 'addressable/uri'
 gem 'bootstrap-sass'
+gem 'flag-icon-sass', :git => 'git@github.com:cfiorini/flag-icon-sass.git', :branch => 'master'
 gem 'delayed_job_active_record'
 gem 'dotenv'
 gem 'haml-rails'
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -5,6 +5,7 @@ gem 'rails', '~> 4.2.1'
 
 gem 'addressable', require: 'addressable/uri'
 gem 'bootstrap-sass'
+gem 'flag-icon-sass', :git => 'git@github.com:cfiorini/flag-icon-sass.git', :branch => 'master'
```
0mega: subscribed for this repo to change the branch to tag as soon as it's released

app/views/releases/show.html.haml
```diff
@@ -4,6 +4,12 @@
   View which releases (merged commits) are pending deployment or have already been deployed.
   %br A release is approved if any associated tickets have been approved in JIRA.
 
+%ul.nav.nav-tabs
+  - Rails.configuration.available_deploy_regions.each do |tab_region|
+    %li{ :class => ('active' if (tab_region == @region)), :role => "presentation" }
```
LogaJ: Tiny nit/comment, seems a pity to add another instance variable to do this check.

I don't have a solution, though...

andistuder:
LGTM

Gemfile
dideler: The latest version is on RubyGems https://rubygems.org/gems/flag-icon-sass/versions/

app/controllers/releases_controller.rb
```diff
@@ -30,7 +30,7 @@ def app_name
   end
 
   def region
-    cookies[:deploy_region]
+    @region = cookies[:deploy_region]
```
dideler: Not a blocker: we may want to start using locals soon instead of passing ivars to the view from all over

app/views/releases/show.html.haml
```diff
@@ -4,6 +4,12 @@
   View which releases (merged commits) are pending deployment or have already been deployed.
   %br A release is approved if any associated tickets have been approved in JIRA.
 
+%ul.nav.nav-tabs
+  - Rails.configuration.available_deploy_regions.each do |tab_region|
```
dideler: Was just updated to `deploy_regions` in #95

app/views/releases/show.html.haml
dideler: Nitpick: prefer new hash syntax for consistency with the rest of the file

dideler:
Looks good! One blocking comment above.

andistuder:
LGTM

PR #97, opened 2016-02-10
Adds slack notifier wrapper to send a message to a specific channel

0mega:
`#send` method might be improved and just receive arguments to fill a template warning message.
E.g.: `#send(feature_review_link, username, date_time, channel)` after we define what should be included in the message.

It's also possible to direct the message to a specific user, instead of sending to a channel.


0mega:
ping @FundingCircle/black-bat

.env.development.example
```diff
@@ -17,3 +17,6 @@ DEPLOY_REGIONS=gb,us
 GITHUB_REPO_STATUS_ACCESS_TOKEN=example_token
 JIRA_HOST_NAME=hostname_of_your_jira_sandbox
 DATA_MAINTENANCE=false
+
+# Slack notifier Web hook
```
dideler: Nitpick: Comment seems redundant, can clarify what it's for in the example value.

lib/slack/notifications.rb
```diff
@@ -0,0 +1,12 @@
+require 'slack-notifier'
+
+class SlackNotifier
+  def initialize
+    @notifier = Slack::Notifier.new ENV['SLACK_WEB_HOOK']
```
dideler: Will sending blow up if the notifier was initialised with `nil`?

andistuder:
LGTM

0mega:
The content of this PR will be part of #99

PR #98, opened 2016-02-15
Feature Review requires reapproval

andistuder:
When a Feature Review is linked to a ticket *after* approval, it will now have a "Requires reapproval" status which is an "unauthorised" (red) status.

This new status can appear on the Releases and Feature Review pages.

On the Releases page, such a release will link to a Feature Review with events projected up until the current time, that is, projection will not be locked to the approval timestamp, since the approval is not legitimate. This also solves the problem where viewing a Feature Review that was linked to a Ticket *after* approval meant you would not see any associated tickets on the Feature Review page, because events would only be projected up until the approval time.

### DIFF
--- a/app/decorators/feature_review_with_statuses.rb
+++ b/app/decorators/feature_review_with_statuses.rb
@@ -62,6 +62,15 @@ class FeatureReviewWithStatuses < SimpleDelegator
     end
   end
 
+  def authorised?
+    return false if tickets.empty?
+
+    tickets.each do |ticket|
+      return false if ticket.approved_at.nil? || ticket.approved_at < ticket.event_created_at
+    end
+    true
+  end
+
   def approved_at
     return unless approved?
     @approved_at ||= tickets.map(&:approved_at).max
--- a/app/decorators/feature_review_with_statuses.rb
+++ b/app/decorators/feature_review_with_statuses.rb
@@ -67,7 +67,7 @@ class FeatureReviewWithStatuses < SimpleDelegator
   end
 
   def authorisation_status
-    return :not_approved unless approved?
+    return :not_approved unless tickets_approved?
 
     if authorised?
       :approved
@@ -77,11 +77,11 @@ class FeatureReviewWithStatuses < SimpleDelegator
   end
 
   def approved_at
-    return unless approved?
+    return unless tickets_approved?
     @approved_at ||= tickets.map(&:approved_at).max
   end
 
-  def approved?
+  def tickets_approved?
     @approved ||= tickets.present? && tickets.all?(&:approved?)
   end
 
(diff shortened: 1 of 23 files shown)


dideler:
@FundingCircle/black-bat need to do a bit more renaming to clarify between approval and authorisation, and make rubocop happy, but you can start reviewing.

app/helpers/feature_reviews_helper.rb
```diff
@@ -44,8 +44,8 @@ def to_link(url, options = {})
   end
 
   def feature_status(feature_review)
-    status = "Feature Status: #{feature_review.approval_status.to_s.humanize}"
-    status << " at #{feature_review.approved_at}" if feature_review.approved? && feature_review.approved_at
+    status = "Feature Status: #{feature_review.authorisation_status.to_s.humanize}"
+    status << " at #{feature_review.approved_at}" if feature_review.authorised? && feature_review.approved_at
```
0mega: testing for `approved_at` may not be needed, as authorised does test for approved_at, via tickets.

andistuder:
LGTM, so far. LMK when you have the authorized/approved renaming done.

dideler:
@andistuder @0mega renaming in code done, will verify if anything in features needs to be renamed

app/decorators/feature_review_with_statuses.rb
```diff
@@ -62,20 +62,30 @@ def summary_status
     end
   end
 
+  def authorised?
```
andistuder: consider calling this `release_authorised?`

app/decorators/feature_review_with_statuses.rb
```diff
@@ -62,20 +62,30 @@ def summary_status
...
+    return :not_approved unless tickets_approved?
+
+    if authorised?
+      :approved
+    else
+      :requires_reapproval
+    end
+  end
+
   def approved_at
```
andistuder: consider calling this `tickets_approved_at`

app/helpers/feature_reviews_helper.rb
dideler: Nice spot, fa2e95c275c52701bff167b8b19b86ae0e7bdb3a

andistuder:
OK, I put a couple of suggestions in. But no blockers. If you consider this ready to merged, please remove the WIP label. LGTM.

app/decorators/feature_review_with_statuses.rb
dideler: I thought about that but kept it as is since it could be confusing as we have Releases, and a Feature Review doesn't know anything about a Release. But a Feature Review does know about its associated tickets, so I updated `approved?` to `tickets_approved?`, which I think should remove any confusion from `authorised?`.

@0mega what do you think?

app/decorators/feature_review_with_statuses.rb
dideler: 727fd6b7969ca3a7f672b4f3fa5e8bb1c09380ad

PR #99, opened 2016-02-15
Audit deploy after processing incoming event

0mega:
So that unauthorised deploys are caught. For now we only check if the deploy version is on master.
If deployed version is not on master we log a warning and send a notification to Slack.


app/models/deploy_alert.rb
```diff
@@ -0,0 +1,31 @@
+require 'repositories/deploy_repository'
+require 'git_repository_loader'
```
dideler: Missing require for `GitRepositoryLocation`

app/models/repositories/deploy_repository.rb
```diff
@@ -40,12 +51,21 @@ def apply(event)
         deployed_by: event.deployed_by,
         event_created_at: event.created_at,
       )
+
+      audit_deploy(Deploy.new(deploy.attributes))
```
dideler: Is creating a Deploy object necessary? Looks like we can just pass `deploy.attributes`; the `auditable?` check is already being performed in `audit`.

lib/slack/notifications.rb
```diff
@@ -0,0 +1,12 @@
+require 'slack-notifier'
+
+class SlackNotifier
+  def initialize
+    @notifier = Slack::Notifier.new(ENV.fetch('SLACK_WEBHOOK'))
```
dideler: If we'll always use a single webhook URI, we could make this a singleton/cache it.

0mega: As this is run from a delayed job, I think it loads the application when spawning a job, so thing caching thing would not happen, it has to recreate the object every time (If I'm not mistaken)

app/models/git_repository.rb
```diff
@@ -103,24 +103,6 @@ def path
...
-    walker.hide(hide_commit_oid)
-    walker
-  end
-
-  def merge_to_master_commit(commit_oid)
-    walker = get_walker(main_branch.target_id, commit_oid, true)
-    walker.find { |commit| rugged_repository.descendant_of?(commit.oid, commit_oid) }
-  end
-
   def commit_on_master?(commit_oid)
```
dideler: Reminder that this will need to be tested.

app/jobs/deploy_alert_job.rb
```diff
@@ -0,0 +1,11 @@
+class DeployAlertJob < ActiveJob::Base
```
dideler:
```ruby
require 'slack/notifications'
```

lib/slack/notifications.rb
dideler: I think DJ's worker process just spawns the app once, but I could be wrong. I experimented with making this is a singleton and queued a job multiple times, and the object ID of the notifier in `#perform` was always the same.

app/jobs/deploy_alert_job.rb
```diff
@@ -0,0 +1,11 @@
...
+  queue_as :default
+
+  def perform(deploy)
+    message = DeployAlert.audit(deploy)
```
andistuder: I reckon at this point we should 'de-serialise' the deploy data. So the interface of `audit` can be changed to expect a `Deploy` object, as was ..

app/models/git_repository.rb
0mega: a simple true/false test done [here](https://github.com/FundingCircle/shipment_tracker/pull/99/files#diff-49b368c6950304cb7e824b84575c80e4R403)

0mega:
ping @FundingCircle/black-bat please review

app/models/repositories/deploy_repository.rb
```diff
@@ -40,12 +52,19 @@ def apply(event)
...
+      audit_deploy(deploy.attributes)
     end
 
     private
 
     attr_reader :store
 
+    def audit_deploy(deploy_attrs)
+      deploy_attrs['event_created_at'] = deploy_attrs['event_created_at'].to_s
+      DeployAlertJob.perform_later(deploy_attrs)
```
dideler: Instead of scheduling a job and possibly aborting it in `DeployAlertJob` if there is no message, suggest checking if it's alertable here before scheduling to prevent the DJ stack potentially taking a hit.

Also suggest taking the `#auditable?` step outside of `#audit` and used that as the condition instead of checking if there is a message.

app/models/repositories/deploy_repository.rb
```diff
@@ -40,12 +52,19 @@ def apply(event)
...
+
+      audit_deploy(deploy.attributes)
     end
 
     private
 
     attr_reader :store
 
+    def audit_deploy(deploy_attrs)
+      deploy_attrs['event_created_at'] = deploy_attrs['event_created_at'].to_s
```
dideler: Is this needed? Won't string interpolation handle this?

spec/jobs/deploy_alert_job_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+        'environment' => 'production',
+        'region' => 'us',
+      }
+    }
+    let(:expected_arg) {
+      Deploy.new(deploy_attrs)
+    }
+
+    it 'runs DeployAlertJob.audit with correct arguments' do
+      allow(DeployAlert).to receive(:audit)
```
dideler: This stub shouldn't be necessary since it's being mocked on the next line.

app/models/repositories/deploy_repository.rb
```diff
@@ -23,6 +25,16 @@ def deploys_for_versions(versions, environment:, region:)
...
 
+    # TODO: will be used for further DeployAlerting
+    # def deploys_ordered_by_id(order:, environment:, region:)
+    #   query = store.select('*')
+    #   query = query.where(environment: environment)
+    #   query = query.where(region: region)
+    #   query.order("id #{order}").map { |deploy_record|
+    #     Deploy.new(deploy_record.attributes)
+    #   }
+    # end
```
dideler: Unstage this little bit?

app/models/repositories/deploy_repository.rb
0mega: Unfortunately it passes a Time object which than raises an exception, given that a job cannot receive complex objects. So I think we have to do this.

spec/models/deploy_alert_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+  let(:staging_env) { 'staging' }
+  let(:production_env) { 'production' }
+
+  describe '.auditable?' do
+    before do
+      allow(GitRepositoryLocation).to receive(:app_names).and_return([app_name])
+    end
+
+    context 'non-production env' do
+      let(:deploy) { Deploy.new(environment: staging_env) }
```
dideler: nitpick: suggest hardcoding the environment here instead of a `let` that's only used here (could also remove production `let`s to make tests more readable)

spec/models/deploy_alert_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    context 'app not under audit' do
+      let(:deploy) { Deploy.new(app_name: 'something', environment: production_env) }
+
+      it 'returns false' do
+        expect(DeployAlert.auditable?(deploy)).to be false
+      end
+    end
+
+    context 'production env and app is under audit' do
+      let(:deploy) { Deploy.new(environment: production_env, app_name: app_name) }
```
dideler: nitpick: attributes are in different order than in above test

spec/models/deploy_alert_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      it 'returns true' do
+        expect(DeployAlert.auditable?(deploy)).to be true
+      end
+    end
+  end
+
+  describe '#audit' do
+    let(:git_diagram) { '-o-A-o' }
```
dideler: Doesn't look like this `let` is being used.

spec/models/deploy_alert_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    let(:repository_loader) { instance_double(GitRepositoryLoader) }
+    let(:test_git_repo) { Support::RepositoryBuilder.build(git_diagram) }
+    let(:rugged_repo) { Rugged::Repository.new(test_git_repo.dir) }
+    let(:repo) { GitRepository.new(rugged_repo) }
+
+    before do
+      allow(GitRepositoryLoader).to receive(:from_rails_config).and_return(repository_loader)
+      allow(repository_loader).to receive(:load).with('frontend').and_return(repo)
```
dideler: Could use `app_name` here.

spec/models/deploy_alert_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    #   xit 'returns true'
+    # end
+
+    # context 'deploy to analyzed environment is not in the commits range' do
+
+    # end
+
+    # context 'last deploy to analyzed environment has unapproved ticket' do
+    #
+    # end
```
dideler: Unstage?

spec/models/git_repository_spec.rb
```diff
@@ -400,6 +400,38 @@
...
+        expect(repo.commit_on_master?(sha)).to be false
+      end
+    end
+
+    context 'the requested commit is on master' do
+      let(:git_diagram) do
+        <<-'EOS'
+          o-A-B---
+         /        \
+       -o-------o--C---o
```
dideler: Is there a simpler diagram we can use? This one indicates that `A` and `B` are also if interest.

spec/models/repositories/deploy_repository_spec.rb
```diff
@@ -13,6 +13,33 @@
     end
   end
 
+  describe '#apply' do
```
dideler: Could we call this something else, e.g. 'schedules an alert job'

It's not really testing the `apply` method (which is moreso being tested in the other tests).

spec/jobs/deploy_alert_job_spec.rb
0mega: Ta.

app/models/repositories/deploy_repository.rb
andistuder: we'd like to leave this in as we'll pick up work on this end in 1-2 weeks.

spec/models/repositories/deploy_repository_spec.rb
andistuder: It tests the aspect of #apply that we can not efficiently test via query.  
The other tests testing the query.

app/models/repositories/deploy_repository.rb
0mega: I understand your concern, but sincerely I think that the number of deploys won't scale that much so that this can become a bottleneck. I imagine at most having `x` number of deploy events at the same moment, being `x` the number of apps one have in production. And this won't happen every minute not even every hour.
Do you agree this can be dealt with in the future if it becomes a problem?

dideler: I think it would also make the code easier to follow if we only schedule when it's auditable, rather than scheduling, checking if it's auditable, then also checking if there's a message. But this small change will require changes to several tests, so up to you if you want to refactor it or not. Not a blocker.

0mega: Ok, I agree, I think I will leave it for the next iteration of this story.

0mega:
@dideler I think everything is addressed, except from starting delayed job conditionally. I think we can do it in the next iteration of this feature, would really like to have this in production to observe the behaviour.

dideler:
LGTM :+1:

andistuder:
to confirm: the nodes are provisioned with the new Slack webhook env var.

PR #100, opened 2016-02-16
Update rake task

0mega:
So that we can pass locale and deployer for testing new features.

- Allows to pass deployer, so that it comes through when testing Slack notifications
- Allows to pass locale, so that we can test the Multi Region filtering

### DIFF
--- a/lib/tasks/send_event.rake
+++ b/lib/tasks/send_event.rake
@@ -14,9 +14,9 @@ namespace :send do
   end
 
   desc 'Sends a sample deploy event'
-  task :deploy_event, [:app_name, :version, :server, :environment, :url] do |_, args|
+  task :deploy_event, [:app_name, :version, :server, :environment, :url, :region, :deployer] do |_, args|
     usage = 'Usage: rake "send:deploy_event[app_name, s0m3c0mm1t, app_name.example.com, production, '\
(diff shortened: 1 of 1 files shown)


dideler:
LGTM, may also want to update script in hello_world_rails

lib/tasks/send_event.rake
```diff
@@ -14,9 +14,9 @@ namespace :send do
   end
 
   desc 'Sends a sample deploy event'
-  task :deploy_event, [:app_name, :version, :server, :environment, :url] do |_, args|
+  task :deploy_event, [:app_name, :version, :server, :environment, :url, :region, :deployer] do |_, args|
     usage = 'Usage: rake "send:deploy_event[app_name, s0m3c0mm1t, app_name.example.com, production, '\
-            'http://shipment_tracker.url/events/deploy?token=s0m3t0k3n]"'
+            'http://shipment_tracker.url/events/deploy?token=s0m3t0k3n, gb, okruk]"'
```
LogaJ: :question: what's the significance of `okruk`?

0mega: If you check the task arguments you can see that the last one is the deployer. I will change it to `a_user` instead

LogaJ: Ah okay that makes sense :+1:

PR #101, opened 2016-02-17
Improve UX of onboarding repositories

dideler:
- [x] Improve copy on page
- [x] Reject empty URIs
- [x] Reject invalid Git URIs
- [x] Reject repositories that are not accessible (does not exist or lacking read permissions)
  - This only works for GitHub repositories; such validation is skipped if they are hosted elsewhere
- [x] Accept all valid variations of Git URIs


dideler:
@FundingCircle/black-bat please review

app/controllers/git_repository_locations_controller.rb
```diff
@@ -1,23 +1,36 @@
 class GitRepositoryLocationsController < ApplicationController
   def index
-    @new_git_repository_location = GitRepositoryLocation.new
+    @repo_location_form = repo_location_form
```
LogaJ: Is this used in a view somewhere? If not I'm wondering why it's needed for the index page at all.

app/models/forms/repository_locations_form.rb
```diff
@@ -0,0 +1,70 @@
...
+  class RepositoryLocationsForm
+    extend ActiveModel::Naming
+    include ActiveModel::Validations
+
+    def to_key
+      nil
+    end
+
+    WHITELIST_DOMAINS = %w(github.com).freeze
+    REPO_URI_REGEX = %r{((file|git|ssh|http(s)?)|(git@[\w\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)?(/)?}
```
LogaJ: Is this an exhaustive list of git repo regexes that can occur?

app/models/forms/repository_locations_form.rb
```diff
@@ -0,0 +1,70 @@
...
+        errors.add(:git_uri, 'cannot be empty')
+        return false
+      end
+
+      unless uri =~ REPO_URI_REGEX
+        errors.add(:git_uri, 'must be valid Git URI, e.g. git@github.com:owner/repo.git')
+        return false
+      end
+
+      unless valid_hostname?
```
LogaJ: Suggestion, you should move this validation out as it was surprising to me that it was here, as the method name mentions only `uri`s and not `hostnames`.

app/models/git_repository_loader.rb
```diff
@@ -59,8 +60,8 @@ def repository_dir_name(git_repository_location)
   end
 
   def options_for(uri, &block)
-    case URI.parse(uri).scheme
-    when 'ssh'
+    parsed_uri = GitCloneUrl.parse(uri)
+    if ['ssh', 'git', nil].include?(parsed_uri.scheme) && parsed_uri.user == 'git'
```
LogaJ: What does this array signify? It is a bit magical.

app/models/git_repository_location.rb
```diff
@@ -47,23 +37,22 @@ def self.update_from_github_notification(payload)
...
-    remote_url
-  end
-
   def extract_name(uri)
     uri.chomp('.git').split('/').last
   end
+
+  def self.url_from_uri(uri)
+    parsed_uri = GitCloneUrl.parse(uri)
+    path = parsed_uri.path.start_with?('/') ? parsed_uri.path[1..-1] : parsed_uri.path
```
LogaJ: `parsed_uri.path[1..-1]` Is a bit fiddly to get my head around. Why would this be the correct path if the `parsed_uri` starts with `/`. Seems a bit magical.

app/models/octokit_client.rb
```diff
@@ -0,0 +1,24 @@
...
+  def initialize(token: ENV['GITHUB_REPO_STATUS_ACCESS_TOKEN'])
+    @client = Octokit::Client.new(access_token: token)
+  end
+
+  def repo_accessible?(uri)
+    parsed_uri = GitCloneUrl.parse(uri)
+    return unless parsed_uri.host == 'github.com'
+
+    path = parsed_uri.path
+    repo_path = path.start_with?('/') ? path[1..-1] : path
```
LogaJ: As above.

features/support/hooks/repo_check_mock.rb
```diff
@@ -0,0 +1,3 @@
+Before('@disable_repo_verification') do
+  allow_any_instance_of(OctokitClient).to receive(:repo_accessible?).and_return(true)
```
LogaJ: `OctoKitClient.repo_accessible?` can sometimes return `nil` but this allow means that never gets tested. I might have missed a test that explicitly does not stub out this instance and passes in values to make `#repo_accessible?`...

app/controllers/git_repository_locations_controller.rb
0mega: It's used here: https://github.com/FundingCircle/shipment_tracker/pull/101/files#diff-7aaa40ddccf3ab52404d0178e6b15304R3

app/models/forms/repository_locations_form.rb
dideler: As far as we know, this will match all valid Git URIs.

app/models/git_repository_loader.rb
dideler: Git URI protocols that we want to use SSH for:
- `ssh://...`
- `git://...`
- `git@...`

app/models/forms/repository_locations_form.rb
0mega: hostname is part of the URI, so I think it should be expected. E.g.: scheme://hostname/path?query_string

app/models/octokit_client.rb
dideler: Octokit requires the repo name to be in the form `owner/repo`, if we give it `/owner/repo` it will fail, so we need to strip the leading `/` if it's present.

app/models/git_repository_location.rb
0mega: This is because the parsed_uri does not always contain `/`. For example when one parses this URI `git@github.com:owner/repo.git`, the path would be `owner/repo.git`. Whereas for `https://github.com/owner/repo.git` the path would be `/owner/repo.git`.
To solve the difference we always remove the prefixed `/` and when the URI is recreated here https://github.com/FundingCircle/shipment_tracker/pull/101/files#diff-32a711be0a73a66e319e654df83ad1fcR55 we always add `/`.

features/support/hooks/repo_check_mock.rb
0mega: Added a test in 82f1479460435016cab8d6545846ea140daca5b8

app/models/git_repository_loader.rb
LogaJ: I think it's the `nil` bit as it doesn't map (in my mind) to any protocol.

0mega: Yes it's strange, but the nil protocol occurs when you have this type of URI: `git@github.com:owner/repo.git`. It's implicit ssh protocol.

app/controllers/git_repository_locations_controller.rb
LogaJ: :+1:

app/models/forms/repository_locations_form.rb
LogaJ: Okay that's a fair point.

app/models/git_repository_loader.rb
LogaJ: Ah right, okay.

LogaJ:
LGTM

PR #103, opened 2016-02-18
update git fetch loop 

andistuder:


andistuder:
@LogaJ as I was cycling home, I was thinking this over a little.. here a suggestion. Best review commit individually to see reason for change.

PR #104, opened 2016-02-25
Token generation

0mega:
This PR enables selecting Tokens to be automatically generated when adding a new repository.

- CircleCI (webhook) and Deployment tokens are enabled by default
- A successful message now appears when user adds a repository


app/models/forms/repository_locations_form.rb
```diff
@@ -14,17 +15,30 @@ def to_key
...
+      @token_types = token_types
     end
 
     def valid?
       valid_uri? && repo_accessible? && errors.empty?
     end
 
+    def self.default_token_types
+      @default_token_types ||= Token.sources.map { |token_src|
+        value = if DEFAULT_SELECTED_TOKENS.include? token_src.endpoint
```
dideler: Can be rewritten as
```ruby
value = DEFAULT_SELECTED_TOKENS.include?(token_src.endpoint)
``

File: app/controllers/git_repository_locations_controller.rb [hunk 2]
```diff
@@ -2,35 +2,41 @@ class GitRepositoryLocationsController < ApplicationController
...
-      render :index
-    end
-  end
+    @token_types = Forms::RepositoryLocationsForm.default_token_types
 
-  private
+    AddRepositoryLocation.run(git_repo_location_params).match do
+      success do |repo_name|
+        flash[:success] = "Successfuly added #{repo_name} repository. " \
+                              'Selected tokens were generated and can be found on Tokens page.'
```
dideler: The token message will show if no tokens were selected?

app/controllers/git_repository_locations_controller.rb
```diff
@@ -2,35 +2,41 @@ class GitRepositoryLocationsController < ApplicationController
...
+      end
 
-  def errors
-    if @git_repository_location.errors.empty?
-      @repo_location_form.errors.full_messages.to_sentence
-    else
-      @git_repository_location.errors.full_messages.to_sentence
+      failure do |errors|
+        @git_repository_locations = GitRepositoryLocation.all.order(:name)
+        flash.now[:error] = errors.errors if errors
```
dideler: `errors.errors if errors` :hushed: 

Can we rename `#errors` or can we do something like `flash.now[:error] = errors` and if `errors` is empty nothing shows?

app/controllers/git_repository_locations_controller.rb
```diff
@@ -2,35 +2,41 @@ class GitRepositoryLocationsController < ApplicationController
...
+      params.dig(:forms_repository_locations_form, :uri),
+      params[:token_types],
+    )
   end
 
-  def git_repository_location_params
-    params.require(:forms_repository_locations_form).permit(:uri)
+  def git_repo_location_params
+    permitted = params.require(:forms_repository_locations_form).permit(:uri)
+    permitted[:token_types] = params[:token_types] if params[:token_types].present?
```
dideler: Maybe we can add more info inside the `forms_repository_locations_form` param hash, then we can do
```ruby
params.require(:forms_repository_locations_form).permit(:uri, :tokens)
```

app/controllers/git_repository_locations_controller.rb
```diff
@@ -2,35 +2,41 @@ class GitRepositoryLocationsController < ApplicationController
...
+    permitted[:validation_form] = @repo_location_form
```
dideler: `@repo_location_form` is more than web param, do we need to include it as part of `git_repo_location_params`?

app/controllers/tokens_controller.rb
```diff
@@ -1,7 +1,7 @@
 class TokensController < ApplicationController
   def index
     @token = Token.new
-    @tokens = Token.all
+    @tokens = Token.order('LOWER(name)')
```
dideler: Can lowercase the name before we save instead of lowercasing on every request.

(Mentioned on the commit, but repeating here so it's discussable in the PR)

app/models/forms/repository_locations_form.rb
```diff
@@ -14,17 +15,30 @@ def to_key
...
     WHITELIST_DOMAINS = %w(github.com).freeze
     REPO_URI_REGEX = %r{((file|git|ssh|http(s)?)|(git@[\w\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)?(/)?}
+    DEFAULT_SELECTED_TOKENS = %w(circleci deploy).freeze
 
     attr_reader :uri
 
-    def initialize(uri)
+    def initialize(uri, token_types)
       @uri = uri
+      @token_types = token_types
```
dideler: Where is this instance variable used?

dideler:
Will review the other half of the PR in the morning :sun_with_face:

app/controllers/tokens_controller.rb
0mega: The only problem is duplication, but it's better then this operation for each row on each request. We can also restrict the names to be lowercase and convert the old values to lowercase, but it's not that nice.

app/controllers/git_repository_locations_controller.rb
0mega: ups... :astonished:

app/controllers/git_repository_locations_controller.rb
0mega: will improve this

app/models/forms/repository_locations_form.rb
0mega: Now it's used in https://github.com/FundingCircle/shipment_tracker/pull/104/files#diff-3bfefd77bbeaf26d56bc08a1919bdee4R32

app/use_cases/add_repository_location.rb
```diff
@@ -0,0 +1,44 @@
...
+  def validate(args)
+    validation_form = args[:validation_form]
+
+    return fail :invalid_uri, errors: get_errors_msg(validation_form) unless validation_form.valid?
+    continue(args)
+  end
+
+  def add_repo(args)
+    uri = args[:uri]
+    git_repo_location = GitRepositoryLocation.new(uri: uri)
```
dideler: Please explicitly require this dependency.

Can also inline the variables that aren't reused, e.g.
```ruby
GitRepositoryLocation.new(uri: args[:uri])
...
args[:repo_name] = git_repo_location.name
```

app/use_cases/add_repository_location.rb
```diff
@@ -0,0 +1,44 @@
...
+      tokens_created &= token.save
+    end
+
+    return fail :failed_generating_token unless tokens_created
+    continue(repo_name)
+  end
+
+  private
+
+  def get_errors_msg(obj)
```
dideler: I like to avoid `get` and `set` method names unless it's to indicate that the cost of the call is expensive.[<sup>1</sup>](https://google.github.io/styleguide/pyguide.html?showone=Access_Control#Access_Control)

Maybe something like `errors_for(obj)`?

app/views/git_repository_locations/index.html.haml
```diff
@@ -6,6 +6,15 @@
     .col-md-4
       = form.text_field(:uri, placeholder: 'git@github.com:user/repo.git', class: 'form-control')
   .form-group.row
+    %label.col-md-1.control-label Tokens
+    .col-md-2
+      - @token_types.each do |token|
+        .checkbox
+          %label
+            = check_box_tag("token_types[#{token[:id]}]", token[:id], token[:value], name: 'token_types[]')
```
dideler: Does the name need to have the `[]`?

spec/models/forms/repository_locations_form_spec.rb
```diff
@@ -1,8 +1,20 @@
-require 'spec_helper'
+require 'rails_helper'
```
dideler: Why was the switch needed?

spec/models/forms/repository_locations_form_spec.rb
```diff
@@ -1,8 +1,20 @@
...
+      { id: 'circleci-manual', name: 'CircleCI (post test)', value: false },
+      { id: 'deploy', name: 'Deployment', value: true },
+      { id: 'jenkins', name: 'Jenkins', value: false },
+      { id: 'jira', name: 'JIRA', value: false },
+      { id: 'uat', name: 'UAT', value: false },
+      { id: 'github_notifications', name: 'Github Notifications', value: false },
+    ]
+  }
+
+  subject(:form) { Forms::RepositoryLocationsForm.new(git_uri, expected_tokens) }
```
dideler: Is it possible to have the `tokens` parameter have a default value of `nil`? Reading the tests, it seems like the tokens affect validation but I don't think they do.

app/controllers/git_repository_locations_controller.rb
```diff
@@ -2,35 +2,48 @@ class GitRepositoryLocationsController < ApplicationController
...
-  def errors
-    if @git_repository_location.errors.empty?
-      @repo_location_form.errors.full_messages.to_sentence
-    else
-      @git_repository_location.errors.full_messages.to_sentence
-    end
+  def get_success_msg(repo_name)
+    msg = "Successfuly added #{repo_name} repository."
+    msg.concat(' Selected tokens were generated'\
+      'and can be found on Tokens page.') unless @repo_location_form.token_types.blank?
```
dideler: Looks like you need a whitespace after "generated" or before "and"

dideler: Is it possible to use a narrower (non-rails) check? e.g. `.empty?`

dideler:
LGTM :+1:  No blockers, just suggestions.

app/views/git_repository_locations/index.html.haml
0mega: Yes, it's a convention. It allows Rails to return all the selected check_boxes with that name in a single array, which is quite handy.

spec/models/forms/repository_locations_form_spec.rb
0mega: We now have a dependency on `Token` model which in turn has a dependency on `ActiveRecord`. Unfortunately mocking the `Token` doesn't avoid the loading of ActiveRecord since the class is still evaluated when creating the mock.

spec/models/forms/repository_locations_form_spec.rb
0mega: yes it can :)

PR #105, opened 2016-02-28
Use a separate token for GitHub repo check

dideler:
The repository validation that happens when onboarding a new repository should prevent the adding of repos that cannot be cloned. This requires a token to be used that has the same permissions as the user who does the cloning.

Related to https://github.com/FundingCircle/shipment_tracker-deployment/pull/18 and https://github.com/FundingCircle/devtools/pull/273


spec/controllers/github_notifications_controller_spec.rb
```diff
@@ -2,97 +2,88 @@
...
 
       context 'when the pull request is newly opened' do
-        let(:sha) { '12345' }
-        let(:repo_url) { 'https://github.com/FundingCircle/hello_world_rails' }
-        let(:payload) { github_pr_payload(action: 'opened', repo_url: repo_url, sha: sha) }
-
-        it 'sets the pull request status' do
-          request.host = 'foo.bar'
-
+        it 'reset and sets the pull request status' do
```
LogaJ: small nit: `resets`

LogaJ:
LGTM! :+1:

0mega:
Nice work :nail_care:, LGTM

PR #106, opened 2016-02-29
Ensure Tempfile object is closed after written to.

LogaJ:
I suspect that's why https://app.honeybadger.io/projects/44071/faults/27840709#notice-summary happened.

I had to use `begin; ensure` because passing in a block doesn't return the file object, which means we can't find out the `#path` that's needed by the `Rugged::Credentials::SsKey` class.


dideler:
:+1:

PR #108, opened 2016-02-29
[ENG-597] Adds a Jira client allowing to post comments to Jira issues

0mega:
This PR creates a wrapper for jira-ruby gem, and allows to post comments on behalf of a user configured through environment variables.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -35,6 +35,4 @@ Rails.application.configure do
   # config.action_view.raise_on_missing_translations = true
 
   config.active_job.queue_adapter = :delayed_job
-
-  config.jira_host_name = ENV.fetch('JIRA_HOST_NAME', 'localhost')
 end
--- a/lib/clients/jira.rb
+++ b/lib/clients/jira.rb
@@ -1,0 +1,29 @@
+require 'jira'
+
+class JiraClient
+  def self.post_comment(issue_id, comment_msg)
+    issue = get_issue(issue_id)
+    comment = issue.comments.build
+    comment.save('body': comment_msg)
+  end
+
+  def self.get_issue(id)
+    client = jira_client
+    client.Issue.find(id)
+  end
+  private_class_method :get_issue
+
+  def self.jira_client
+    options = {
+      username: ENV['JIRA_USER'],
+      password: ENV['JIRA_PASSWD'],
+      site: ENV['JIRA_HOST'],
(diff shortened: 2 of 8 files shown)


lib/clients/jira.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class JiraClient
+  def self.post_comment(issue_id, comment_msg)
+    issue = get_issue(issue_id)
+    comment = issue.comments.build
+    comment.save('body': comment_msg)
+  end
+
+  def self.get_issue(id)
+    client = jira_client
```
dideler: nitpick: this assignment doesn't seem necessary

lib/clients/jira.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def self.jira_client
+    options = {
+      username: ENV['JIRA_USER'],
+      password: ENV['JIRA_PASSWD'],
+      site: ENV['JIRA_HOST'],
+      context_path: ENV['JIRA_PATH'],
+      auth_type: :basic,
+      read_timeout: 120,
+    }
```
dideler: minor: options can also be cached to avoid recreating every time the client is used

lib/clients/jira.rb
```diff
@@ -0,0 +1,29 @@
...
+    client.Issue.find(id)
+  end
+  private_class_method :get_issue
+
+  def self.jira_client
+    options = {
+      username: ENV['JIRA_USER'],
+      password: ENV['JIRA_PASSWD'],
+      site: ENV['JIRA_HOST'],
+      context_path: ENV['JIRA_PATH'],
```
dideler: nitpick: do you think it's worth setting these as constants in `application.rb`?

dideler:
:+1: 

We should add tests at some point, but not a blocker now since the class isn't being used yet.

config/environments/development.rb
```diff
@@ -35,6 +35,4 @@
   # config.action_view.raise_on_missing_translations = true
 
   config.active_job.queue_adapter = :delayed_job
-
-  config.jira_host_name = ENV.fetch('JIRA_HOST_NAME', 'localhost')
```
dideler: Since this had a fallback, we should ensure we have the value in our `.env.development`

0mega: Agreed, although it's very unlikely you will be running Jira on localhost.

andistuder:
LGTM

PR #111, opened 2016-03-01
Improve git loop performance

andistuder:
With this change, we only update a repo if the remote head has been updated.

### DIFF
--- a/app/models/git_repository_location.rb
+++ b/app/models/git_repository_location.rb
@@ -32,6 +32,10 @@ class GitRepositoryLocation < ActiveRecord::Base
     find_by('uri LIKE ?', "%#{repo_name}%")
   end
 
+  def self.app_remote_head_hash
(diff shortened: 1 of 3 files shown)


andistuder:
@dideler @LogaJ @0mega I did a short investigation on the performance optimisation failed and here a fix. Note the actual change in in this commit: https://github.com/FundingCircle/shipment_tracker/commit/da86a0d68e422caa51e68626490db6c3eb9d7f44

dideler:
If we can prove on UAT that this works and is a noticeable speed improvement then :+1: from me.

LogaJ:
didto @dideler's comment.

dideler:
Ideally we extract the logic out of the rake task like we did for the event updater, that way we can also test it since it's an important flow and will make future changes safer to make.

0mega:
LGTM

PR #113, opened 2016-03-01
Update env vars when booting locally with heroku settings

dideler:
- Add `GITHUB_REPO_READ_TOKEN` for repo validation

- Add `GITHUB_REPO_STATUS_WRITE_TOKEN` for posting commit statuses

- Use foreman long options for readability


LogaJ:
Could you also add the JIRA settings currently in our .env.development file?

dideler:
@LogaJ discussed with @0mega and we'll wait until our JIRA work is integrated to see if it's needed in development.

0mega:
:+1:

andistuder:
Great! Jira infrastructure is in place, so you should be able to include that now.

dideler:
Added JIRA settings plus the option to override any of the values (e.g. if you have a JIRA running locally and would rather use that instance).

LogaJ:
:+1:

PR #114, opened 2016-03-01
Auto-link Feature Review to JIRA ticket if previous link exists

dideler:
As a developer
When I have linked a Feature Review to a JIRA ticket
And I create a new Feature Review for the same branch
I want the new FR to be automatically linked to the existing ticket


app/controllers/github_notifications_controller.rb
```diff
@@ -1,16 +1,14 @@
...
 require 'payloads/github'
-require 'pull_request_status'
 
 class GithubNotificationsController < ApplicationController
   skip_before_action :verify_authenticity_token
 
   def create
     if pull_request?
-      process_pull_request
+      HandlePullRequestEvent.run(payload)
```
andistuder: nice extraction!

app/use_cases/handle_push_event.rb
```diff
@@ -0,0 +1,43 @@
...
+    git_repository_location = GitRepositoryLocation.find_by_full_repo_name(payload.full_repo_name)
+    return fail :repo_not_found unless git_repository_location
+    git_repository_location.update(remote_head: payload.after_sha)
+    continue(payload)
+  end
+
+  def relink_tickets(payload)
+    ticket_repo = Repositories::TicketRepository.new
+    linked_tickets = ticket_repo.tickets_for_versions(payload.before_sha)
+    return fail :no_linked_tickets if linked_tickets.empty?
```
andistuder: do you really want to `fail`? I assume that will prevent any further steps from being run.

dideler: We decided to keep the fail and return early. If a future business requirement causes the behaviour to change, it'll either be a separate use case or we'll modify this one not to fail.

Semantically, we consider it a failure of the step if tickets were not linked. (Note that solid-use-case's `fail` has different behaviour from ruby's builtin `fail, it doesn't raise and it should be used to tell the caller which step has not completed).

app/jobs/relink_ticket_job.rb
```diff
@@ -0,0 +1,39 @@
+require 'clients/jira'
+require 'honeybadger'
+require 'repositories/ticket_repository'
+
+class RelinkTicketJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(payload)
```
andistuder: maybe call this `args`  instead of `payload` - I find it confusing as elsewhere the `payload` is a GitHub payload object

dideler: Good point.

app/jobs/relink_ticket_job.rb
```diff
@@ -0,0 +1,39 @@
...
+    JiraClient.post_comment(ticket_key, message)
+  rescue JiraClient::InvalidKeyError
+    Rails.logger.error "Failed to post comment to Jira ticket #{ticket_key}."\
+      'The issue might have been deleted'
+  rescue StandardError => error
+    Honeybadger.notify(error)
+  end
+
+  def feature_review_url(feature_review_path)
+    Rails.application.routes.url_helpers.root_url.chomp('/') + feature_review_path
```
andistuder: is `app.root_url` not sufficient?

dideler: Thanks, will try.

app/jobs/relink_ticket_job.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  private
+
+  def link_feature_review_to_ticket(ticket_key, old_feature_review_path, before_sha, after_sha)
+    new_feature_review_path = old_feature_review_path.sub(before_sha, after_sha)
+    message = "[Feature ready for review|#{feature_review_url(new_feature_review_path)}]"
+    JiraClient.post_comment(ticket_key, message)
+  rescue JiraClient::InvalidKeyError
+    Rails.logger.error "Failed to post comment to JIRA ticket #{ticket_key}. Ticket might have been deleted."
```
andistuder: I think log level `warn` is sufficient.

app/jobs/relink_ticket_job.rb
dideler: Doesn't seem to work, will keep as is.

app/jobs/relink_ticket_job.rb
dideler: :+1:

app/use_cases/handle_push_event.rb
```diff
@@ -0,0 +1,20 @@
+require 'git_repository_location'
+require 'relink_ticket_job'
+
+class HandlePushEvent
+  include SolidUseCase
+
+  steps :update_remote_head, :relink_tickets
```
andistuder: maybe start with a validate step here?
eg. check if app under audit.. instead of doing it inside the update remote head step

config/application.rb
```diff
@@ -35,7 +35,7 @@ class Application < Rails::Application
 
     routes.default_url_options = {
       protocol: ENV.fetch('PROTOCOL', 'https'),
-      host: ENV.fetch('HOST_NAME', "localhost:#{ENV.fetch('PORT', '1201')}"),
+      host: ENV.fetch('HOST_NAME', ENV['PORT'] ? "localhost:#{ENV['PORT']}" : 'localhost'),
```
andistuder: OK

spec/support/feature_review_helpers.rb
```diff
@@ -25,7 +25,7 @@ def self.build(*args)
         new.build(*args)
       end
 
-      def initialize(host = 'http://test.host')
+      def initialize(host = 'https://localhost')
```
andistuder: why changing this to localhost?

andistuder:
LGMT - only question to investigate is the GitHub push notification payload where SHAs may not be SHAs..

dideler:
@andistuder re: "removed" shas, so far we only seem to get these for PullRequestEvents. Let's investigate a bit more tomorrow, but it may not be an issue for PushEvents. Hoping to hear back from GitHub Support by then.

app/jobs/relink_ticket_job.rb
```diff
@@ -0,0 +1,38 @@
...
+
+class RelinkTicketJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(args)
+    before_sha = args.delete(:before_sha)
+    after_sha = args.delete(:after_sha)
+
+    ticket_repo = Repositories::TicketRepository.new
+    linked_tickets = ticket_repo.tickets_for_versions(before_sha)
```
andistuder: Does this work? I thought `tickets_for_versions` expects an array - It might, as we do a type conversion in the DB query. If it works, please extend the test in ticket_repository_spec.rb to document the behaviour with a string.. to change to `[before_sha]`.

app/jobs/relink_ticket_job.rb
andistuder: :+1:

app/use_cases/handle_push_event.rb
dideler: We've added a validate step in the other PR :+1:

app/jobs/relink_ticket_job.rb
dideler: Nice catch, didn't realize we were doing it. Turns out it does work!
```
Repositories::TicketRepository.new.tickets_for_versions 'c44386b6f69bbbbd73fa315417ca4bd122b6e279'
  Snapshots::Ticket Load (3.2ms)  SELECT DISTINCT ON (key) * FROM "tickets" WHERE (versions && ARRAY['c44386b6f69bbbbd73fa315417ca4bd122b6e279']::varchar[])  ORDER BY key, id DESC
=> [#<Ticket key="ENG-286" summary="quick tickets" status="Ready for Acceptance" paths=["/feature_reviews?apps%5Bhello_world_rails%5D=c44386b6f69bbbbd73fa315417ca4bd122b6e279"] approved_at=nil version_timestamps={"c44386b6f69bbbbd73fa315417ca4bd122b6e279"=>Mon, 22 Feb 2016 14:17:48 +0000} versions=["c44386b6f69bbbbd73fa315417ca4bd122b6e279"]>, #<Ticket key="ENG-292" summary="Sample story" status="Ready for Acceptance" paths=["/feature_reviews?apps%5Bhello_world_rails%5D=c44386b6f69bbbbd73fa315417ca4bd122b6e279"] approved_at=nil version_timestamps={"c44386b6f69bbbbd73fa315417ca4bd122b6e279"=>Mon, 22 Feb 2016 14:17:13 +0000} versions=["c44386b6f69bbbbd73fa315417ca4bd122b6e279"]>]
```

Will add a test.

PR #116, opened 2016-03-02
Use jira Fully Qualified Domain Name instead of just domain name

0mega:
So that protocol is explicit for communication between Jira and ST


0mega:
@FundingCircle/black-bat :eyes:

dideler:
:+1:

PR #117, opened 2016-03-02
Link Jira issue via form

andistuder:
allows linking a Feature Review to a Jira issue via web form.


andistuder:
@0mega @dideler  please review

.env.development.example
```diff
@@ -23,3 +23,6 @@ DATA_MAINTENANCE=false
 SLACK_WEBHOOK=slack_notifier_webhook
 
 ALLOW_GIT_FETCH_ON_REQUEST=true
+
+PROTOCOL=http
+PORT=1201
```
0mega: not clear `PROTOCOL` and `PORT` for what? Could we have this prefixed with the service name or some context?

bin/boot_with_heroku_settings
```diff
@@ -5,7 +5,7 @@ SSH_PRIVATE_KEY=${SSH_PRIVATE_KEY:-$(heroku config:get SSH_PRIVATE_KEY)} \
 SSH_PUBLIC_KEY=${SSH_PUBLIC_KEY:-$(heroku config:get SSH_PUBLIC_KEY)} \
 GITHUB_REPO_READ_TOKEN=${GITHUB_REPO_READ_TOKEN:-$(heroku config:get GITHUB_REPO_READ_TOKEN)} \
 GITHUB_REPO_STATUS_WRITE_TOKEN=${GITHUB_REPO_STATUS_WRITE_TOKEN:-$(heroku config:get GITHUB_REPO_STATUS_WRITE_TOKEN)} \
-JIRA_HOST_NAME=${JIRA_HOST_NAME:-$(heroku config:get JIRA_HOST_NAME)} \
+JIRA_FQDN=${JIRA_FQDN:-$(heroku config:get JIRA_FQDN)} \
```
0mega: Thanks

config/routes.rb
```diff
@@ -13,6 +13,9 @@
   # Projections
   resource :feature_reviews, only: [:new, :show, :create] do
     get 'search'
+    member do
+      post 'link_ticket'
+    end
```
0mega: Should we really keep this? If it does not change anything, I'd prefer not to include it, otherwise in future we might think it does some magic that we just can't find.

app/controllers/feature_reviews_controller.rb
```diff
@@ -62,4 +70,38 @@ def feature_review_form
...
+  def redirect_path
+    @redirect_path ||= path_from_url(params[:return_to])
+  end
+
+  def post_jira_comment
+    JiraClient.post_comment(jira_key, jira_comment) if valid_format?(jira_key)
+    flash[:success] = "Feature Review was linked to #{jira_key}."\
+    ' Refresh this page in a moment and the ticket will appear.'
+  rescue StandardError => error
+    if error.is_a?(JiraClient::InvalidKeyError)
```
dideler: Prefer to have multiple rescues, one for `JiraClient::InvalidKeyError` and one for anything else.

app/controllers/feature_reviews_controller.rb
```diff
@@ -62,4 +70,38 @@ def feature_review_form
...
+      Honeybadger.notify(error)
+    end
+  end
+
+  def jira_comment
+    "[Feature ready for review|#{feature_review_url}]"
+  end
+
+  def valid_format?(jira_key)
+    fail JiraClient::InvalidKeyError unless /[A-Z][A-Z]+-\d*/ =~ jira_key
```
dideler: Where is this exception defined?

app/controllers/feature_reviews_controller.rb
```diff
@@ -62,4 +70,38 @@ def feature_review_form
   def git_repository_for(app_name)
     git_repository_loader.load(app_name)
   end
+
+  def redirect_path
+    @redirect_path ||= path_from_url(params[:return_to])
+  end
+
+  def post_jira_comment
+    JiraClient.post_comment(jira_key, jira_comment) if valid_format?(jira_key)
```
dideler: Suggestion: consider putting the condition outside of the method, e.g.
```ruby
post_jira_comment if valid_format?
```

That way the method name will accurately reflect its behaviour.

app/controllers/feature_reviews_controller.rb
```diff
@@ -62,4 +70,38 @@ def feature_review_form
...
+    "[Feature ready for review|#{feature_review_url}]"
+  end
+
+  def valid_format?(jira_key)
+    fail JiraClient::InvalidKeyError unless /[A-Z][A-Z]+-\d*/ =~ jira_key
+    true
+  end
+
+  def jira_key
+    request.request_parameters['jira_key']
```
dideler: Is it not possible to get this from Rails `params`?

.env.development.example
andistuder: it's for application itself - PORT already existed: https://github.com/FundingCircle/shipment_tracker/blob/jira-ticket-via-form/config/application.rb#L36

app/controllers/feature_reviews_controller.rb
andistuder: yes, but as I would like to (re) use the error rescue (L:83), I would need to introduce a further level of nesting.

app/controllers/feature_reviews_controller.rb
andistuder: Good point, this has moved on a little since I first wrote it. I will refactor.

config/routes.rb
andistuder: I'll remove it.

app/helpers/application_helper.rb
```diff
@@ -7,7 +7,6 @@ def wiki_links(key)
       qa: 'https://github.com/FundingCircle/shipment_tracker/wiki/3.-QA-REVIEWS-FEATURE',
       releases: 'https://github.com/FundingCircle/shipment_tracker/wiki/5.-MERGING-AND-DEPLOYING-A-FEATURE',
       approve: 'https://github.com/FundingCircle/shipment_tracker/wiki/4.-PRODUCT-OWNER-REVIEWS-FEATURE',
-      adding_tickets: 'https://github.com/FundingCircle/shipment_tracker/wiki/1.-CREATING-A-FEATURE-REVIEW#adding-jira-tickets-to-the-feature-review',
```
dideler: How come this is removed? The old way is still supported, and we can update the documentation to also mention the new way.

config/application.rb
```diff
@@ -33,7 +33,10 @@ class Application < Rails::Application
 
     config.action_dispatch.perform_deep_munge = false
 
-    routes.default_url_options = { host: ENV['HOST_NAME'] || "localhost:#{ENV['PORT']}" }
+    routes.default_url_options = {
+      protocol: ENV['PROTOCOL'] || 'https',
+      host: ENV['HOST_NAME'] || "localhost:#{ENV['PORT']}",
```
dideler: Prefer fetch with fallback.

app/controllers/feature_reviews_controller.rb
andistuder: yes, good point. thanks.

app/helpers/application_helper.rb
andistuder: Personally, I don't think the new flow needs help. But will check with Steve. Beyond, I agree that we should keep info on manual linking in the wiki - so, I won't delete that section, but amend it.

dideler:
LGTM :+1:

config/application.rb
andistuder: will do.

PR #119, opened 2016-03-03
Allow usage of short SHA version

0mega:
So that it's still possible to query repositories using the short version, 7 characters.


dideler:
LGTM :+1:

PR #120, opened 2016-03-04
Improve Status notification behaviour

0mega:
Curretly we have the status update reporting an error on github status while the feature is being relink.
This PR avoids this intermediate error step which makes it better for usability.


app/jobs/relink_ticket_job.rb
```diff
@@ -18,10 +18,17 @@ def perform(args)
...
     end
+
+    update_status(args)
   end
 
   private
 
+  def update_status(args)
+    status_options = { full_repo_name: args[:full_repo_name], sha: args[:head_sha] }
+    PullRequestUpdateJob.set(wait: 10.seconds).perform_later(status_options)
```
andistuder: :no_entry_sign: please only update the GitHub status if you do not link the feature review. When linking a feature review, we will get a webhook notification from Jira, which will then update the status.

andistuder:
`HandlePullRequestEvent` seems no longer used. Remove class?

app/use_cases/handle_push_event.rb
```diff
@@ -4,17 +4,34 @@
...
+    # Only enables the functionality for our test repo
+    # TODO: remove when production ready
+    return fail :other_repos unless payload.full_repo_name == 'FundingCircle/hello_world_rails'
     git_repository_location = GitRepositoryLocation.find_by_full_repo_name(payload.full_repo_name)
     return fail :repo_not_found unless git_repository_location
     git_repository_location.update(remote_head: payload.after_sha)
     continue(payload)
   end
 
+  def reset_commit_status(payload)
```
andistuder: is this really needed? With this we actually introduce some flip-flop behaviour. Can we not just wait for the updates to propagate? Via delayed job, the status will be either set to error (if no relinking taks place, or success if relinking does take place, via Jira)

app/use_cases/handle_push_event.rb
```diff
@@ -4,17 +4,34 @@
 class HandlePushEvent
   include SolidUseCase
 
-  steps :update_remote_head, :relink_tickets
+  steps :update_remote_head, :reset_commit_status, :relink_tickets
 
   def update_remote_head(payload)
+    # Only enables the functionality for our test repo
+    # TODO: remove when production ready
+    return fail :other_repos unless payload.full_repo_name == 'FundingCircle/hello_world_rails'
```
andistuder: good security gate - let's remember to remove it.

app/models/pull_request_status.rb
```diff
@@ -19,26 +19,33 @@ def update(full_repo_name:, sha:)
...
-    github.create_status(
-      repo: full_repo_name,
-      sha: sha,
-      state: searching_status[:status],
-      description: searching_status[:description],
-    )
+    post_status(full_repo_name, sha, searching_status)
+  end
+
+  def error(full_repo_name:, sha:)
```
andistuder: is this method used?

app/models/pull_request_status.rb
```diff
@@ -19,26 +19,33 @@ def update(full_repo_name:, sha:)
...
+    post_status(full_repo_name, sha, error_status)
+  end
+
+  def not_found(full_repo_name:, sha:)
```
andistuder: is this method used?

app/jobs/relink_ticket_job.rb
dideler: We realized the same thing and have already changed it. We'll push our latest changes now.

dideler:
Some of the feedback is already addressed in unpushed work. We'll push the latest now.

app/use_cases/handle_push_event.rb
dideler: Keeping this ensures that there is instant feedback (i.e. a searching status) whenever a developer pushes. I don't see any strong reason to remove it. There won't be a race condition with the final status because that will always occur afterwards.

app/models/pull_request_status.rb
dideler: Not yet, but we will use it in the `RelinkTicketJob` when an unexpected error occurs, to avoid an everlasting "searching" status.

lib/clients/github.rb
```diff
@@ -13,6 +13,8 @@ def create_status(repo:, sha:, state:, description:, target_url: nil)
       description: description,
       target_url: target_url
     )
+  rescue Octokit::NotFound
```
andistuder: :+1: for rescue error.

dideler:
@andistuder please re-review

app/models/pull_request_status.rb
dideler: Yup, we post a "not found" failure status when relinking tickets but there are no tickets to relink.

app/jobs/relink_ticket_job.rb
```diff
@@ -8,6 +8,8 @@ class RelinkTicketJob < ActiveJob::Base
   def perform(args)
     before_sha = args.delete(:before_sha)
     after_sha = args.delete(:after_sha)
+    full_repo_name = args.delete(:full_repo_name)
+    head_sha = args.delete(:head_sha)
 
     ticket_repo = Repositories::TicketRepository.new
     linked_tickets = ticket_repo.tickets_for_versions(before_sha)
```
andistuder: Does `tickets_for_versions` work correctly when a single version is passed in? I think we only test passing in of a Array.

app/jobs/relink_ticket_job.rb
```diff
@@ -27,12 +32,22 @@ def link_feature_review_to_ticket(ticket_key, old_feature_review_path, before_sh
     message = "[Feature ready for review|#{feature_review_url(new_feature_review_path)}]"
     JiraClient.post_comment(ticket_key, message)
   rescue JiraClient::InvalidKeyError
+    @send_error_status = true
```
andistuder: :+1:

spec/models/payloads/github_spec.rb
```diff
@@ -41,7 +41,10 @@
   describe '#head_sha' do
     context 'with pull_request data' do
       it 'returns sha' do
-        data = { 'pull_request' => { 'head' => { 'sha' => 'abc123' } } }
+        data = { 'head_commit' => {
```
andistuder: I would prefer if we could have a full payload here - maybe introduce it as fixture. (Not a blocker...)

spec/models/payloads/github_spec.rb
```diff
@@ -61,12 +64,8 @@
     context 'with pull_request data' do
       it 'returns html url' do
         data = {
-          'pull_request' => {
-            'base' => {
-              'repo' => {
-                'html_url' => 'https://github.com/foo/bar',
-              },
-            },
+          'repository' => {
```
andistuder: as above, a full payload would give me more confidence that this test correctly mocks the interface with GitHub (not a blocker)

andistuder:
LGTM.. Let's merge this asap, so we can prepare approval of the parent branch. I move the comment on tickets_for_versions into there..

app/jobs/relink_ticket_job.rb
andistuder: ignore this, I raised this in the parent PR.. where the change actually is introduced.

app/models/pull_request_status.rb
andistuder: :ok_hand:

andistuder:
No need to have a Shipment Tracker Feature review as this is not a PR to master ;)

PR #121, opened 2016-03-04
send deploy alert if deploy version unknown

andistuder:
Given we receive production deploy events with version no given (i.e. NULL), we want to raise a deploy alert, so that we can establish what the deployer has deployed.

### DIFF
--- a/app/models/deploy_alert.rb
+++ b/app/models/deploy_alert.rb
@@ -14,7 +14,7 @@ class DeployAlert
 
     github_repo = GitRepositoryLoader.from_rails_config.load(deploy.app_name)
 
-    alert_not_on_master(deploy) unless github_repo.commit_on_master?(deploy.version)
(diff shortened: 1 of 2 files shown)


app/models/deploy_alert.rb
```diff
@@ -25,6 +25,6 @@ def self.audit(deploy)
   def self.alert_not_on_master(deploy)
     time = deploy.event_created_at.strftime('%F %H:%M%:z')
     "Deploy Alert for #{deploy.app_name} at #{time}. #{deploy.deployed_by} " \
-    "deployed version #{deploy.version} not on master branch."
+    "deployed version #{deploy.version || 'unkown'} not on master branch."
```
dideler: typo "unknown"

andistuder: un-what?! .. thanks for spotting!

dideler:
:+1:

PR #123, opened 2016-03-07
fix bug when posting too many statuses

andistuder:
After receiving a GitHub push notification, we seem to update the GitHub commit status for all versions linked to the related tickets. This seems wasteful. I reckon we should only update the version mentioned in the Push notification as after. Needs a bit more investigation.


dideler:
Could you explain the bug? Do we have a JIRA issue for it?

PR #126, opened 2016-03-08
Update deploy alert message

0mega:
- Updates alert message
- Makes slack channel for notifications configurable
- Adds deployment verification rules:
  - Alert if a deploy contains at least one version which was not approved


spec/models/deploy_alert_spec.rb
```diff
@@ -77,12 +77,13 @@
...
           version: sha, environment: 'production', app_name: app_name,
-          deployed_by: 'user1', event_created_at: DateTime.parse('2016-02-15T15:57:25+01:00'))
+          deployed_by: 'user1', event_created_at: DateTime.parse('2016-02-15T15:57:25+01:00'),
+          region: 'uk')
       }
 
       let(:expected_message) {
-        'Deploy Alert for frontend at 2016-02-15 15:57+01:00. ' \
-        "user1 deployed version #{sha} not on master branch."
+        'UK Deploy Alert for frontend at 2016-02-15 15:57+01:00. ' \
```
andistuder: Nitpick: We never do UK, it's not in ISO 3166-1 alpha-2.

andistuder:
LGTM

config/application.rb
```diff
@@ -14,6 +14,7 @@
 Dotenv.load
 
 module ShipmentTracker
+  DEPLOY_ALERT_SLACK_CHANNEL ||= ENV.fetch('DEPLOY_ALERT_SLACK_CHANNEL', 'general')
```
andistuder: Note that the env variable does not contain `#` anymore. 
https://github.com/FundingCircle/shipment_tracker-deployment/commit/a4afebe9a72901de80b6c9e9110be23d2d0ae251
Does the client actually expect it?

app/models/repositories/deploy_repository.rb
```diff
@@ -25,15 +25,15 @@ def deploys_for_versions(versions, environment:, region:)
...
-    # TODO: will be used for further DeployAlerting
-    # def deploys_ordered_by_id(order:, environment:, region:)
-    #   query = store.select('*')
-    #   query = query.where(environment: environment)
-    #   query = query.where(region: region)
-    #   query.order("id #{order}").map { |deploy_record|
-    #     Deploy.new(deploy_record.attributes)
-    #   }
-    # end
+    def deploys_ordered_by_id(order:, environment:, region:, limit: nil)
```
andistuder: I think you may want to query for `app_name` also. 

Further, please ensure that this query performs: use `explain` to check if/what DB index it uses. Maybe you would find it easier to make it perform well if you don't allow any conditionals. Maybe tailor it exactly to the envisaged use (I assume it is to get the 'old_deploy')... eg. call it `current_production_deploy_for_region` and allow `region` and `app_name` as arguments only.

app/jobs/deploy_alert_job.rb
```diff
@@ -6,10 +6,14 @@ class DeployAlertJob < ActiveJob::Base
...
-    deploy = Deploy.new(deploy_attrs)
-    message = DeployAlert.audit(deploy)
+    new_deploy = Deploy.new(deploy_attrs[:new_deploy])
+    old_deploy = Deploy.new(deploy_attrs[:old_deploy]) if deploy_attrs[:old_deploy]
+
+    message = DeployAlert.audit(new_deploy, old_deploy)
+
     return unless message
+
     Rails.logger.warn message
```
dideler: Now that we are actually sending the message, the log call can be removed.

app/models/repositories/deploy_repository.rb
0mega: Nice catch. We added a regression test for making sure app name is considered.

Also the query was refactored and analysis looks good. The execution cost is low. I can share details later.

app/models/deploy_alert.rb
```diff
@@ -5,26 +5,75 @@
...
 
-  def self.auditable?(deploy)
-    deploy.environment == 'production' && GitRepositoryLocation.app_names.include?(deploy.app_name)
+  def self.auditable?(new_deploy)
+    new_deploy.environment == 'production' && GitRepositoryLocation.app_names.include?(new_deploy.app_name)
   end
 
-  def self.audit(deploy)
-    return unless auditable?(deploy)
+  def self.audit(new_deploy, previous_deploy = nil)
```
dideler: Minor: Consider rename to `audit_message` since that's what it returns

app/models/deploy_alert.rb
```diff
@@ -5,26 +5,75 @@
...
-    # @deploy_repo = Repositories::DeployRepository.new
-    # deploy_region = deploy_vo.region
-    # deploys = @deploy_repo.deploys_ordered_by_id('desc', environment: deploy_env, region: deploy_region)
-    # last_deploy, before_last_deploy = deploys[0..1]
+    alert_not_authorised(new_deploy) unless deploy_auditor.all_releases_authorised?
+  end
+
+  def self.alert_not_authorised(deploy)
+    time = deploy.event_created_at.strftime('%F %H:%M%:z')
+    "#{deploy.region&.upcase} Deploy Alert for #{deploy.app_name} at #{time}. #{deploy.deployed_by} " \
```
dideler: Does Slack allow newlines in its messages? If so, consider putting a newline after
`"<Region> Deploy Alert for <app> at <time>"`

config/application.rb
0mega: Now it does ;)

app/models/deploy_alert.rb
```diff
@@ -5,26 +5,75 @@
...
+    "deployed #{deploy.version || 'unknown'}, release not authorised."
   end
 
   def self.alert_not_on_master(deploy)
     time = deploy.event_created_at.strftime('%F %H:%M%:z')
-    "Deploy Alert for #{deploy.app_name} at #{time}. #{deploy.deployed_by} " \
-    "deployed version #{deploy.version || 'unknown'} not on master branch."
+    "#{deploy.region&.upcase} Deploy Alert for #{deploy.app_name} at #{time}. #{deploy.deployed_by} " \
+    "deployed #{deploy.version || 'unknown'} not on master branch."
```
dideler: Minor: I would prefer if an alert caused by an unknown version uses a different message than when the version is not on master. The current message is misleading if a software version that exists on master was actually deployed, but due to some error the version we receive is nil.

app/models/deploy_alert.rb
```diff
@@ -5,26 +5,75 @@
...
+      @new_deploy = new_deploy
+      @previous_deploy = previous_deploy
+    end
+
+    def unknown_or_not_on_master?
+      @new_deploy.version.nil? || !@github_repo.commit_on_master?(@new_deploy.version)
+    end
+
+    def all_releases_authorised?
+      auditable_commits = auditable_commits_list
```
dideler: `#auditable_commits_list` could be renamed to `#auditable_commits` and the list can be cached.

app/models/deploy_alert.rb
```diff
@@ -5,26 +5,75 @@
...
-  def self.auditable?(deploy)
-    deploy.environment == 'production' && GitRepositoryLocation.app_names.include?(deploy.app_name)
+  def self.auditable?(new_deploy)
+    new_deploy.environment == 'production' && GitRepositoryLocation.app_names.include?(new_deploy.app_name)
   end
 
-  def self.audit(deploy)
-    return unless auditable?(deploy)
+  def self.audit(new_deploy, previous_deploy = nil)
+    github_repo = GitRepositoryLoader.from_rails_config.load(new_deploy.app_name)
```
dideler: Looks like the github_repo is only used inside the DeployAuditor. Maybe pass DeployAuditor the app name instead and have it create the github_repo?

app/models/deploy_alert.rb
```diff
@@ -5,26 +5,75 @@
...
+        git_repo: github_repo,
+        app_name: app_name,
+        commits: auditable_commits,
+      )
+    end
+
+    def auditable_commits_list
+      return [@github_repo.commit_for_version(@new_deploy.version)] unless @previous_deploy
+
+      @github_repo.commits_between(@previous_deploy.version, @new_deploy.version, true)
```
dideler: Could we make the last parameter a keyword argument? Makes it easier to tell what is being flagged when the method is called.

app/models/deploy_alert.rb
```diff
@@ -5,26 +5,75 @@
...
+      release_query = release_query_for(
+        auditable_commits,
+        @new_deploy.region,
+        @github_repo,
+        @new_deploy.app_name,
+      )
+
+      auditable_releases = release_query.deployed_releases
+
+      auditable_releases.all?(&:authorised?)
```
dideler: Suggestion: Can inline and avoid the variable assignment, still reads nice.

app/models/deploy_alert.rb
```diff
@@ -5,26 +5,75 @@
...
+      @github_repo = github_repo
+      @new_deploy = new_deploy
+      @previous_deploy = previous_deploy
+    end
+
+    def unknown_or_not_on_master?
+      @new_deploy.version.nil? || !@github_repo.commit_on_master?(@new_deploy.version)
+    end
+
+    def all_releases_authorised?
```
dideler: Is there a better name we can use here? E.g. `recent_releases_authorised?` as it's not checking all releases.

app/models/git_repository.rb
```diff
@@ -21,16 +21,12 @@ def exists?(full_sha)
     false
   end
 
-  def commits_between(from, to)
+  def commits_between(from, to, simplified = false)
```
dideler: Minor: In `def get_walker` we use `simplify`, nicer if consistently named (I don't have a preference which one is used)

app/models/git_repository.rb
```diff
@@ -21,16 +21,12 @@ def exists?(full_sha)
...
     instrument('commits_between') do
       validate_commit!(from) unless from.nil?
       validate_commit!(to)
 
-      walker = Rugged::Walker.new(rugged_repository)
-      walker.sorting(Rugged::SORT_TOPO | Rugged::SORT_REVERSE) # optional
-      walker.push(to)
-      walker.hide(from) if from
```
dideler: `#get_walker` does not conditionally hide, is there a case where we pass it a nil `from` commit?

app/models/git_repository.rb
```diff
@@ -44,6 +40,10 @@ def recent_commits_on_main_branch(count = 50)
     build_commits(walker.take(count))
   end
 
+  def commit_for_version(sha)
```
dideler: Is this method tested?

app/models/queries/releases_query.rb
```diff
@@ -4,7 +4,8 @@ module Queries
   class ReleasesQuery
     attr_reader :pending_releases, :deployed_releases
 
-    def initialize(per_page:, region:, git_repo:, app_name:)
+    def initialize(per_page:, region:, git_repo:, app_name:, commits: nil)
+      @commits = commits if commits
```
dideler: Condition isn't needed.

app/models/repositories/deploy_repository.rb
```diff
@@ -25,16 +25,6 @@ def deploys_for_versions(versions, environment:, region:)
...
-    # TODO: will be used for further DeployAlerting
-    # def deploys_ordered_by_id(order:, environment:, region:)
-    #   query = store.select('*')
-    #   query = query.where(environment: environment)
-    #   query = query.where(region: region)
-    #   query.order("id #{order}").map { |deploy_record|
-    #     Deploy.new(deploy_record.attributes)
-    #   }
-    # end
-
```
dideler: :smile:

app/models/repositories/deploy_repository.rb
```diff
@@ -42,8 +32,28 @@ def last_staging_deploy_for_version(version)
...
+           .offset(1)
+           .first
+    end
+
+    private
 
-      deploy = store.create!(
+    attr_reader :store
+
+    def create_deploy_event(event)
```
dideler: This method does not create a deploy event. Maybe call it something like `snapshot_deploy` or `create_deploy_snapshot`?

app/models/repositories/deploy_repository.rb
```diff
@@ -42,8 +32,28 @@ def last_staging_deploy_for_version(version)
...
+      new_deploy = create_deploy_event(event)
+
+      if DeployAlert.auditable?(new_deploy) && !Rails.configuration.data_maintenance_mode
+        previous_deploy = second_last_production_deploy(new_deploy.app_name, new_deploy.region)
+        audit_deploy(new_deploy: new_deploy.attributes, previous_deploy: previous_deploy&.attributes)
+      end
+    end
+
+    def second_last_production_deploy(app_name, region)
+      store.where(app_name: app_name, environment: 'production', region: region)
```
dideler: Should be able to use a `find_by` with multiple conditions to avoid the `#first` call later.

app/models/repositories/deploy_repository.rb
```diff
@@ -52,17 +62,16 @@ def apply(event)
...
+      deploy_time_to_s(deploys_attrs[:new_deploy])
+      deploy_time_to_s(deploys_attrs[:previous_deploy])
+      DeployAlertJob.perform_later(deploys_attrs)
+    end
 
-    def audit_deploy(deploy_attrs)
-      deploy_attrs['event_created_at'] = deploy_attrs['event_created_at'].to_s
-      DeployAlertJob.perform_later(deploy_attrs)
+    def deploy_time_to_s(deploy_attrs)
+      deploy_attrs['event_created_at'] = deploy_attrs['event_created_at'].to_s if deploy_attrs
```
dideler: Could this logic be pushed down to the class that needs it?

bin/boot_with_heroku_settings
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+DEPLOY_ALERT_SLACK_CHANNEL=${DEPLOY_ALERT_SLACK_CHANNEL:-$(heroku config:get DEPLOY_ALERT_SLACK_CHANNEL)} \
```
dideler: Won't we also need the `SLACK_WEBHOOK` env var if we want to test on staging?

features/step_definitions/releases_steps.rb
```diff
@@ -39,3 +39,12 @@
   actual_releases = releases_page.public_send("#{deploy_status}_releases".to_sym)
   expect(actual_releases).to match_array(expected_releases)
 end
+
+Then 'a deploy alert should be dispatched' do
+  expect(SlackNotifier).to have_received(:send)
+    .with(
+      'GB Deploy Alert for frontend at 2016-01-22 17:34+00:00.'\
+      " Jeff deployed #{scenario_context.resolve_version('#merge1')}, release not authorised.",
+      'general',
```
dideler: I'm guessing Slack does accept channel names without `#` prepended?

lib/slack/notifications.rb
```diff
@@ -3,7 +3,13 @@
 class SlackNotifier
   def self.send(msg, channel)
     @notifier ||= Slack::Notifier.new(ENV.fetch('SLACK_WEBHOOK'))
-    @notifier.channel = channel
+    @notifier.channel prepend_hash(channel)
```
dideler: Missing assignment?

app/models/repositories/deploy_repository.rb
0mega: If I remember correctly this is needed for a successful serialisation of the job arguments, so it has to happen before we perform the job.

app/models/repositories/deploy_repository.rb
0mega: We can't use find_by because the criteria won't return one entry, it will return a list of deploys. We also won't be able to easily order the rows.

app/models/git_repository.rb
0mega: Now it is :)

0mega:
@FundingCircle/black-bat this is ready for deploying. Please give it a last quick review.

dideler:
LGTM

app/jobs/deploy_alert_job.rb
```diff
@@ -6,10 +6,13 @@ class DeployAlertJob < ActiveJob::Base
   queue_as :default
 
   def perform(deploy_attrs)
-    deploy = Deploy.new(deploy_attrs)
-    message = DeployAlert.audit(deploy)
+    new_deploy = Deploy.new(deploy_attrs[:new_deploy])
+    old_deploy = Deploy.new(deploy_attrs[:old_deploy]) if deploy_attrs[:old_deploy]
+
+    message = DeployAlert.audit_message(new_deploy, old_deploy)
```
andistuder: Is there a chance that you get a `NameError: undefined local variable or method `old_deploy' ` if `deploy_attrs[:old_deploy]` not present?

andistuder: ignore, I just manually tested this - it seems to work.

app/models/git_repository.rb
```diff
@@ -21,16 +21,12 @@ def exists?(full_sha)
...
     instrument('commits_between') do
       validate_commit!(from) unless from.nil?
       validate_commit!(to)
 
-      walker = Rugged::Walker.new(rugged_repository)
-      walker.sorting(Rugged::SORT_TOPO | Rugged::SORT_REVERSE) # optional
-      walker.push(to)
-      walker.hide(from) if from
-
+      walker = get_walker(to, from, simplify)
```
andistuder: nice tidy up work here! :)

app/models/repositories/deploy_repository.rb
```diff
@@ -42,8 +32,28 @@ def last_staging_deploy_for_version(version)
...
       return unless event.is_a?(Events::DeployEvent)
+      new_deploy = create_deploy_snapshot(event)
+
+      if DeployAlert.auditable?(new_deploy) && !Rails.configuration.data_maintenance_mode
+        previous_deploy = second_last_production_deploy(new_deploy.app_name, new_deploy.region)
+        audit_deploy(new_deploy: new_deploy.attributes, previous_deploy: previous_deploy&.attributes)
+      end
+    end
+
+    def second_last_production_deploy(app_name, region)
```
andistuder: Nitpik: please leave `#apply` as last public method. So the pattern is 1. initialise, 2. queries, 3. appy. .. for all repositories.

app/models/repositories/deploy_repository.rb
```diff
@@ -42,8 +32,28 @@ def last_staging_deploy_for_version(version)
...
+           .first
+    end
+
+    private
 
-      deploy = store.create!(
+    attr_reader :store
+
+    def create_deploy_snapshot(event)
+      store.create(
```
andistuder: :no_entry_sign: you changed to `.create` without '!'. please keep `.create!`,

app/models/repositories/deploy_repository.rb
```diff
@@ -52,17 +62,16 @@ def apply(event)
...
       )
-
-      audit_deploy(deploy.attributes) unless Rails.configuration.data_maintenance_mode
     end
 
-    private
-
-    attr_reader :store
+    def audit_deploy(deploys_attrs)
+      deploy_time_to_s(deploys_attrs[:new_deploy])
```
andistuder: How about pushing this into a method on `Deploy`? eg. `serializable_attrs` - just a suggestion, not a blocker.

app/models/deploy_alert.rb
```diff
@@ -5,26 +5,75 @@
...
+      @new_deploy = new_deploy
+      @previous_deploy = previous_deploy
+    end
+
+    def unknown_or_not_on_master?
+      @new_deploy.version.nil? || !@git_repo.commit_on_master?(@new_deploy.version)
+    end
+
+    def all_releases_authorised?
+      return false unless auditable_commits
```
andistuder: Why is this false? In which business case does auditable_commits returns nil?

app/models/repositories/deploy_repository.rb
0mega: Would look cleaner, but would have to initialize two Deploy objects, then get the serializable_attrs, then recreate the Deploy objects inside DeployAlertJob. So overall, didn't look much better to me. Unless you are talking about Snapshots::Deploy, but that's a bigger refactor I guess.

PR #127, opened 2016-03-08
Fix Releases show requests where app name contains a request format

dideler:
Rails magic was causing requests where the repository name included a valid format (e.g. 'reponame.js') to error.

### DIFF
--- a/app/controllers/releases_controller.rb
+++ b/app/controllers/releases_controller.rb
@@ -1,4 +1,6 @@
 class ReleasesController < ApplicationController
+  before_action :force_html_format, only: :show
+
   def index
     @app_names = GitRepositoryLocation.app_names
   end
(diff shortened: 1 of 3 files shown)


app/controllers/releases_controller.rb
```diff
@@ -26,7 +28,11 @@ def build_projection
   end
 
   def app_name
-    params[:id]
+    if request.path_parameters[:format]
```
LogaJ: You could one line this with:
"#{params[:id]}.#{params.path_parameters.fetch(:format,'')}".chomp('.')

app/controllers/releases_controller.rb
```diff
@@ -26,7 +28,11 @@ def build_projection
...
+      "#{params[:id]}.#{request.path_parameters[:format]}"
+    else
```
LogaJ: You could one line this with:
`"#{params[:id]}.#{params.path_parameters.fetch(:format,'')}".chomp('.')`

0mega:
:+1:

app/controllers/releases_controller.rb
dideler: We prefer to keep it as is, a bit easier to read.

PR #128, opened 2016-03-08
Disable deploy alerts during maintenance mode

dideler:
- [x] Do not resend deploy alerts when in maintenance mode (i.e. recreating snapshots)
- [x] New rake task `recreate_snapshots` which does not process new events received during maintenance mode
- [x] Update docs


0mega:
LGTM

LogaJ:
LGTM

PR #129, opened 2016-03-09
Improve validation when onboarding repositories

dideler:
Git URI validation improvements.

- [x] Strip whitespace from URI
  - Allows valid repos that are padded with whitespace to still be added
- [x] Checks that the regex match occurs at the beginning of the string
  - Avoids URIs like `foo<uri>` from passing regex validation
- [x] Consider repo not accessible if Octokit does not receive correct format for a repository name
  - Prevents app from blowing up if a valid Git URI is not a valid GitHub URI  
     (GitHub has its own restrictions)
- [x] Handle ports in URIs
  - Prevents URIs like `ssh://git@github.com:not_a_port/repo.git` from passing validation
- [x] Improve user feedback

### DIFF
--- a/app/controllers/git_repository_locations_controller.rb
+++ b/app/controllers/git_repository_locations_controller.rb
@@ -29,6 +29,7 @@ class GitRepositoryLocationsController < ApplicationController
 
   def git_repo_location_params
     permitted = params.require(:repository_locations_form).permit(:uri)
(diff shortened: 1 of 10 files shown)


0mega:
LGTM

LogaJ:
LGTM

PR #130, opened 2016-03-11
Send alert if deploy causes production rollback

dideler:
Dependent on PR #126 since it's branched off of it.


app/models/deploy_alert.rb
```diff
@@ -10,54 +10,68 @@ def self.auditable?(new_deploy)
...
-    alert_not_authorised(new_deploy) unless deploy_auditor.all_releases_authorised?
+    deploy_auditor = DeployAuditor.new(new_deploy, previous_deploy)
+
+    if deploy_auditor.unknown_version?
+      alert_unknown_version(new_deploy)
+    elsif deploy_auditor.not_on_master?
+      alert_not_on_master(new_deploy)
+    elsif !deploy_auditor.recent_releases_authorised?
+      alert_not_authorised(new_deploy)
+    end
```
0mega: Much better :)

dideler:
Created #132 which points to master instead.

PR #133, opened 2016-03-14
Do not handle push events for annotated tags

dideler:
Avoids attempting to post commit statuses for annotated tags, which does not work.

Pushes for lightweight tags are still handled, because the SHA will refer to the head commit.
Whereas for annotated tags, the SHA refers to the annotated tag.


andistuder:
:+1:

PR #134, opened 2016-03-15
Freeze all strings

dideler:
@FundingCircle/black-bat this is what it looks like freezing all strings (with rubocop autocorrect).

The build is currently broken because we have code where we mutate string. Will defer fixing the build until we agree to move forward with freezing strings. In terms of benefits, immutable strings are more performant (not sure if we'll notice a difference though) and it eases the transition to Ruby 3.


0mega:
I believe it would be good to fix this :)

0mega:
@FundingCircle/black-bat please review

LogaJ:
LGTM

PR #135, opened 2016-03-15
Experiment: full text search w/ pg_search gem

dideler:
_This PR is not intended to be merged. It's for sharing notes while experimenting with postgres full-text search via the `pg_search` gem. Indexing and fuzzy search are not included in experiment._

---

pg_search supports two different techniques for searching, **multi-search** and **search scopes**.
- Multi-search allows records of many different Active Record classes to be mixed together into one global search index.
- Search scopes allow you to do more advanced searching against only one Active Record class.
  - [Searching associations is supported](https://github.com/Casecommons/pg_search#searching-through-associations) but not recommend because searches cannot be sped up with database indexes. It is supported as a quick way to try out cross-model searching. If you need to search through associations, prefer multi-search over search scopes.

If we end up storing all information into a single table, we will use search scopes, otherwise multi-search.

--- 

pg_search provides Active Record callbacks (only for multi-search?) to keep indexes updated. Some AR operations skip callbacks, so better to have trigger on the DB.

---

A full-text search query for "todo" won't include tickets with status "To Do". We would probably need to [enable fuzzy search](https://github.com/Casecommons/pg_search/wiki/Installing-Postgres-Contrib-Modules). It's a more expensive query, so by default a full-text search should not query with fuzzy matching and fallback to fuzzy search if the first query didn't return relevant results.

---

https://github.com/Casecommons/pg_search#dynamic-search-scopes

[Dynamic search scopes](https://github.com/Casecommons/pg_search#dynamic-search-scopes) allow you to narrow the search scope dynamically.

For example, a search scope against multiple columns on a table would normally look like
```ruby
Snapshots::Something.search_for("whatever")
```

But if we allow special syntax in our search form, we can dynamically narrow scope. E.g.
Given the query string "version:abc123", our search can be narrowed down to a single column
```ruby
Snapshots::Something.search_for(:version, "abc123")
```

---

pg_search uses default full-text search, but you can [search using other features](https://github.com/Casecommons/pg_search#searching-using-different-search-features), e.g.  Trigram search or Double Metaphone search.

https://github.com/Casecommons/pg_search#trigram-trigram-search

Trigram search might be of interest, as it works with typos and misspellings in the query or text.

---

https://github.com/Casecommons/pg_search#weighting

[Each searchable column can be given a weight](https://github.com/Casecommons/pg_search#weighting) of "A", "B", "C", or "D", with "A" being the highest weight (most important). Weights determine the **order** of the returned records. Without query limits, the results will be the same with or without weights, just in different order. When you limit the number of records (and you should), weights will play an important role.

Weighting columns is very simple. In our AR model:
```ruby
# without weighting
pg_search_scope :search_for, against: [:summary, :status]

# with weighting
pg_search_scope :search_for, against: {
  summary: 'A',
  status: 'D'
}
```

Because adding weights is so trivial, I suggest we **start without weighting**. Weights are an optimization, fine tuning the results. We should collect data about searching first, weighting from the start is a premature optimization as we don't yet know what should be weighted and how.

---

https://github.com/Casecommons/pg_search#prefix-postgresql-84-and-newer-only

Searching for partial words is supported (full-text search only). Note this is not the same as fuzzy search or misspelling search. Searching for partial words does a prefix search instead of the default whole word search.

For example, "ENG" will find results that contain "ENG" but will not find results that contain "EG" (requires fuzzy search) or "EGN" (requires misspelling search).

```ruby
WEIGHTED_COLUMNS = { summary: 'A', status: 'D' }
pg_search_scope :search_for, against: WEIGHTED_COLUMNS, using: { tsearch: { prefix: true } }
```

With dynamic scopes, prefix search should be able to be turned on for specific cases (e.g. "ticket:ENG").

---

https://github.com/Casecommons/pg_search#negation

If negation is turned on for full-text search
```ruby
using: { tsearch: { negation: true} }
```

then search terms can easily be excluded from the results, e.g.
```ruby
Snapshots::Something.search_for("release !shipment_tracker")
```

---

https://github.com/Casecommons/pg_search#normalization

[Dictionaries](http://www.postgresql.org/docs/9.2/static/textsearch-dictionaries.html) eliminate words that should not be considered in a search (stop words) and normalize (aka stem) words so that different derived forms of the same word will match.

For example, the "english" dictionary will normalize "jumping" and "jumped" to a lexeme like "jump".

```ruby
using: { tsearch: { dictionary: 'english' } }
```

The default dictionary is "simple", which does not do any normalizing.

--- 

https://github.com/Casecommons/pg_search#normalization

A longer document has a greater chance of containing a query term, so it's reasonable to take into account document size, e.g., a hundred-word document with five instances of a search word is probably more relevant than a thousand-word document with five instances.

The `:normalization` option specifies whether and how a document's length should impact its rank.
```ruby
using: { tsearch: { normalization: 2 } }
```

---

https://github.com/Casecommons/pg_search#any_word

The `:any_word` option returns multiple records when parts of the search term match across different records.

By default, the following would return a match if all of "one two three" is present in a document.
```ruby
Snapshots::Something.search_for("one two three")
```
With `any_word`, it could for example return 3 records if each record contained one of the words.

---

https://github.com/Casecommons/pg_search#ignoring-accent-marks-postgresql-90-and-newer-only

Ignoring accents, nice to have if we get users with multi-language keyboards for example.

---

Boosting performance with tsvector columns

https://github.com/Casecommons/pg_search#using-tsvector-columns
and
https://github.com/Casecommons/pg_search/wiki/Building-indexes#with-activerecord-andor-rails

The `:against` column would only be used when the tsvector_column is not present for the search type.


dideler:
See #138

PR #137, opened 2016-03-15
Use SQL to create tsvector for full text search

LogaJ:
This PR is really for comments only, so that we can discuss implementation of using pure sql to implement full text search.

#135 uses a gem to achieve most (and probably more) of what this PR does.


LogaJ:
Closing now as we've decided which direction to go in. with PR: #136

PR #140, opened 2016-03-16
Search for ticket releases by app name

dideler:
As a user
I want to search for a released ticket* by app name
Because I want to see what features have been released for a specific app

<sup>*A released ticket is a (JIRA) ticket containing one or more Feature Reviews with a deploy for any of the software versions under review.</sup>

Branched off of #136; need to change target branch to master.


db/migrate/20160315165607_add_tsvector_to_tickets.rb
```diff
@@ -27,7 +29,7 @@ def down
     remove_column :released_tickets, :tsv
 
     execute <<-SQL
-      DROP TRIGGER released_tickets_tsv_update
+      DROP TRIGGER IF EXISTS released_tickets_tsv_update
```
andistuder: good spot, thanks!

db/migrate/20160316154428_add_deploys_to_released_tickets.rb
```diff
@@ -0,0 +1,39 @@
...
+          -- setweight(to_tsvector(coalesce(my_function(new.deploys), '')), 'A') ||
+          -- setweight(to_tsvector(coalesce(new.deployed_apps, '')), 'A') ||
+          setweight(to_tsvector(coalesce(new.deploys->>0, '')), 'A') ||
+          setweight(to_tsvector(coalesce(new.summary, '')), 'B') ||
+          setweight(to_tsvector(coalesce(new.description, '')), 'D');
+        return new;
+      end
+      $$ LANGUAGE plpgsql;
+    SQL
+  end
```
andistuder: :no_entry_sign:  don't forget to update the tsv column by triggering an update for each row in the table. as in previous migration: 
```
now = Time.current.to_s(:db)
update("UPDATE released_tickets SET updated_at = '#{now}'")
```

db/migrate/20160316154428_add_deploys_to_released_tickets.rb
```diff
@@ -0,0 +1,39 @@
...
+      CREATE OR REPLACE FUNCTION released_tickets_trigger() RETURNS trigger AS $$
+      begin
+        new.tsv :=
+          setweight(to_tsvector('pg_catalog.english', coalesce(new.summary, '')), 'A') ||
+          setweight(to_tsvector('pg_catalog.english', coalesce(new.description, '')), 'D');
+        return new;
+      end
+      $$ LANGUAGE plpgsql;
+    SQL
+  end
```
andistuder: as above, update for each row to trigger update the tsv row.

db/migrate/20160316154428_add_deploys_to_released_tickets.rb
dideler: We didn't add it because we thought the `update_at` will be updated anyway as a result of the trigger modifications (as the `tsv` column of each row is being modified). But just tested this and doesn't seem to be the case, will update.

dideler: Won't this mass update only happen once, when the migration runs?
Wouldn't it be better to update the time whenever re-indexing happens?

lib/tasks/db.rake
```diff
@@ -1,12 +1,36 @@
...
+    Dotenv.load
+    config_dir = File.expand_path('../../config', File.dirname(__FILE__))
+    file_contents = File.read("#{config_dir}/database.yml.erb")
+
+    File.open("#{config_dir}/database.yml", 'w') do |f|
+      f.write ERB.new(file_contents).result
+    end
+  end
+
+  desc 'reindex released tickets if indexing trigger exists'
```
andistuder: I was thinking this over a little: I think we might actually never need this, as any schema change on a snapshots table is followed by running he recreate snapshots task. Let's discuss...

app/models/repositories/released_ticket_repository.rb
```diff
@@ -1,44 +1,87 @@
...
     end
 
     private
 
-    attr_reader :store, :git_repository_location, :feature_review_factory
+    attr_reader :store, :feature_review_factory
+
+    def snapshot_jira_event(event)
+      feature_reviews = feature_review_factory.create_from_text(event.comment)
+      return if feature_reviews.empty?
```
dideler: Handle the case where edit (e.g. title) happens after FR linked.

app/models/repositories/released_ticket_repository.rb
```diff
@@ -1,44 +1,87 @@
...
+      else
+        store.create!(build_ticket({}, event, feature_reviews))
+      end
+    end
 
-    def previous_ticket_data(key)
-      store.where(key: key).last.try(:attributes) || {}
+    def snapshot_deploy_event(event)
+      tickets_for_version(event.version).each do |record|
+        old_deploys = record.attributes.fetch('deploys', [])
```
dideler: Fallback not needed. Prefer `[]`.

app/models/ticket.rb
```diff
@@ -12,6 +12,7 @@ class Ticket
     attribute :approved_at, DateTime
     attribute :version_timestamps, Hash[String => DateTime]
     attribute :versions, Array, default: []
+    attribute :deploys, Array, default: []
```
dideler: Consider creating separate Virtus object for `ReleasedTicket`

app/views/home/dashboard.html.haml
```diff
@@ -24,3 +24,5 @@
               %p
               %p.key
                 = jira_link(ticket.key)
+              - ticket.deploys.each do |deploy|
+                %p.deploy= "#{deploy['deployed_at']} #{deploy['app']}"
```
dideler: Add link to GitHub commit.

features/dashboard.feature
```diff
@@ -1,19 +1,29 @@
...
-  As a user,
-  I want to free text search for tickets,
-  so I can find tickets related to any topic.
+Scenario: User finds deployed tickets by description
+  Given the following tickets are created:
+    | Jira Key | Summary     | Description | Deploys                     |
+    | ENG-1 | Make this task | As a User\n implement the task | 2016-03-21 12:02 app_1 #abc |
+    | ENG-2 | Make another task | As a User\n implement another task | 2016-03-21 12:02 app_1 #abc |
+    | ENG-3 | Make another story | As a User\n implement another story | 2016-03-21 12:02 app_1 #abc |
+    | ENG-4 | Make another story | As a User\n implement another story |  |
```
dideler: Please align columns.

features/step_definitions/dashboard_steps.rb
```diff
@@ -2,7 +2,13 @@
...
 end
 
-Then 'I should find the following ticket on the dashboard:' do |tickets_table|
+Then 'I should find the following tickets on the dashboard:' do |tickets_table|
   result_tickets = dashboard_page.results
-  expect(result_tickets).to eq tickets_table.hashes
+  hashes = tickets_table.hashes
+
+  hashes.each do |hash|
+    hash['Deploys'] = [hash['Deploys']]
```
dideler: Can use a `split` here.

features/support/sections/results_section.rb
```diff
@@ -21,5 +22,9 @@ def items
     def jira_element(panel_element, klass)
       panel_element.first(".#{klass}")
     end
+
+    def jira_elements(panel_element, klass)
```
dideler: Consider renaming `klass` to `css_class`

lib/tasks/db.rake
dideler: Let's remove it in a new commit so it's still in history.

lib/tasks/jobs.rake
```diff
@@ -41,10 +41,9 @@ namespace :jobs do
 
       updater.run(repo_event_id_hash)
       puts "[#{Time.current}] Completed recreate_snapshots for #{repos.map(&:table_name)}"
-
     }
 
-    t2=Thread.new{
+    t2 = Thread.new{
```
dideler: Can only create one new thread.

andistuder: We decided not to change the threading behaviour here. Main concern is that when running code in the main thread and this code fails, then the other thread(s) will get killed. We'd need to further test if ActiveRecord transaction handles the emitted kill signal correctly. 

If we want to do this, we should do so in a dedicated issue - but only if we have memory usage concerns.

features/step_definitions/dashboard_steps.rb
andistuder: done

features/support/sections/results_section.rb
andistuder: done

app/views/home/dashboard.html.haml
andistuder: Done

app/models/ticket.rb
andistuder: Decided against this until there's a big enough need to create a new Virtus object.

app/models/repositories/released_ticket_repository.rb
0mega: improved here 33864e2510f20f9f3f6fc21194493ad6d0b8a520

features/step_definitions/dashboard_steps.rb
```diff
@@ -7,7 +7,14 @@
   hashes = tickets_table.hashes
 
   hashes.each do |hash|
-    hash['Deploys'] = hash['Deploys'].split(',').map(&:strip)
+    hash['Deploys'] = hash['Deploys'].split(',').map { |deploy_string|
+      deploy_string_without_version, pretend_version = deploy_string.split('#')
+
+      real_version = scenario_context.resolve_version("##{pretend_version}")
```
LogaJ: As a side, this would break if the deploys no longer have the version at the end of the string.

Maybe we can tighten this up by not splitting, but by using a regex instead.

PR #141, opened 2016-03-16
Revert threaded re-snapshoting

0mega:
For some reason this would hand the test suite. Reverting to original non threaded re-snapshoting


dideler:
The revert looks correct, but please verify that it was the threading that caused issues.

If you want, can also revert this extra puts that was added in with the threading: https://github.com/FundingCircle/shipment_tracker/blob/fix-merge/app/models/repositories/updater.rb#L18

0mega:
@dideler not 100% sure but might be related to this issue https://github.com/rails/rails/issues/15089.
Cucumber tests invoke the recreate_snapshots task. I tried setting the `config.allow_concurrency = false` but it doesn't solve the problem.
Without threads everything runs fine. It looks like a testing environment issue only, but it probably doesn't make sense to have one version for testing and another for production, cause then the tests are useless.

andistuder:
LGTM - only see here an alternative fix https://github.com/FundingCircle/shipment_tracker/pull/143 to get the master branch back to :green_apple:  - as explained there the failing tests are a result of a database cleaning strategy incompatible with threaded code. @dideler I hope this addresses https://github.com/FundingCircle/shipment_tracker/pull/141#issuecomment-197456838

But, I agree with @0mega that threading the re-snapshotting is a relatively minor (and optional) improvement and I am ok to go ahead with either of the two fixes (#141 or #143). Your choice..  :)

0mega:
Closing this PR, #143 was chosen as best fix option.

PR #143, opened 2016-03-16
update database cleaner config to fix cukes

andistuder:
As described [here](https://github.com/cucumber/cucumber-rails/blob/master/features/choose_javascript_database_strategy.feature) threading causes issue if the DatabaseCleaner strategy is set to `transaction`. The author suggests to use `truncation` instead, but when running tests with `truncation` other scenarios fail in our test suite. 

The DatabaseCleaner [documentation](https://github.com/DatabaseCleaner/database_cleaner) suggests `deletion` as alternative strategy. This works, but I experienced some flaky test results - until I added an `Around` hook to explicitly clean the Database for each scenario - see [env.rb](https://github.com/FundingCircle/shipment_tracker/compare/fix-merge-database-cleaner?expand=1#diff-57013262a6cdf1f0f6cb2c64aa762c68R69)

### DIFF
--- a/lib/tasks/jobs.rake
+++ b/lib/tasks/jobs.rake
@@ -29,10 +29,10 @@ namespace :jobs do
     manage_pid pid_path_for('jobs_recreate_snapshots')
 
     repos = Rails.configuration.repositories.dup
-    released_tickets_repo = repos.select {|repo| repo.table_name == 'released_tickets'}.first
+    released_tickets_repo = repos.find { |repo| repo.table_name == 'released_tickets' }
     repos.delete(released_tickets_repo)
 
-    t1=Thread.new{
+    t1 = Thread.new{
       puts "[#{Time.current}] Running recreate_snapshots for #{repos.map(&:table_name)}"
       updater = Repositories::Updater.new(repos)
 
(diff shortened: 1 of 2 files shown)


andistuder:
Changing the database cleaner strategy seems to increase the runtime of the cucumber tests by aprox. 10% (was ~11sec, is ~12sec on my machine)

0mega:
This looks good. Is there any impact on tests speed (to worse)? If not I'm glad to use this version over #141

andistuder:
@0mega see comment on speed above - we crossed in the wire :)

lib/tasks/jobs.rake
```diff
@@ -29,10 +29,10 @@ namespace :jobs do
     manage_pid pid_path_for('jobs_recreate_snapshots')
 
     repos = Rails.configuration.repositories.dup
-    released_tickets_repo = repos.select {|repo| repo.table_name == 'released_tickets'}.first
+    released_tickets_repo = repos.find { |repo| repo.table_name == 'released_tickets' }
```
andistuder: wow, check what Rubocop autocorrect produces here! Soon, it will write entire features for us ;)

0mega:
Hehe funny, ok. Will discuss this tomorrow, but we might drop this one due to the test time increase. Although it's not significant on its own, the tests run much more frequently than the rake task, so I'm not sure if we are gaining time by having the threads in the rake task.

lib/tasks/jobs.rake
0mega: :open_mouth: some nice rules they have in there

LogaJ:
LGTM

0mega:
Merging this fix in favor of #141. We agreed to keep in mind this change so that if we notice flaky test we may still go for #141.

PR #144, opened 2016-03-17
Set word-brake to 'break-word' for bootstrap panel-body

0mega:
So that URLs can be broken into several lines when not fitting the panel


LogaJ:
LGTM

dideler:
:+1:

PR #148, opened 2016-03-22
Frontend love

andistuder:
Improve frontend experience for users by providing the following:

- [x] Make title bold.
- [x] Have Jira key in title and linkable.
- [x] Limit length of description maybe 5 lines or so and add ellipsis if it has been truncated.
- [x] Add header to deploys section of ticket and improve format. Include ruler line to break this up.
- [x] Add configurable limit for number of tickets displayed in search results.
- [x] Greyed out default ticket description if non exists, 'No ticket description'.
- [x] Limit title length to 100 characters, with '...'.
- [x] Include info if the ticket has not been deployed 'Deploys - none'
- [x] Change title to 'Feature Reviewed Tickets' and remove the subtitle. 
- [x] Change placeholder text to: 'Search for app name, title, or description'.


app/views/home/dashboard.html.haml
```diff
@@ -18,13 +18,16 @@
...
+            .panel-heading.summary
+              = jira_link(ticket.key)
+              %strong= ticket.summary
             .panel-body
-              = simple_format(ticket.description, class: 'description')
+              = simple_format(truncate(ticket.description, length: 250, seperator: ' '), class: 'description')
               %p
-              %p.key
-                = jira_link(ticket.key)
+              - if !ticket.deploys.empty?
```
andistuder: if `any?`, or unless `empty?`

andistuder:
LGTM - pending PO visual sign-off and manual assertion that db query with limit performs well.

config/application.rb
```diff
@@ -23,6 +23,7 @@ module ShipmentTracker
   GITHUB_REPO_STATUS_WRITE_TOKEN ||= ENV.fetch('GITHUB_REPO_STATUS_WRITE_TOKEN', nil)
   DEFAULT_DEPLOY_LOCALE ||= ENV.fetch('DEFAULT_DEPLOY_LOCALE', 'gb') # For older events without locale
   DEFAULT_HEROKU_DEPLOY_LOCALE ||= ENV.fetch('DEFAULT_HEROKU_DEPLOY_LOCALE', 'us') # When locale not prefixed
+  NUMBER_OF_TICKETS_TO_DISPLAY ||= ENV.fetch('NUMBER_OF_TICKETS_TO_DISPLAY', 100)
```
dideler: Double check with @steveweet how many tickets we should show. I personally find this too high since most people won't look past the first 10 results or so. On the other hand, without pagination, can argue that we need some replacement to show many tickets. But even with pagination, I don't think many people would look past the first page. If they do not find what they're looking for, they'll refine their query and search again.

dideler:
LGTM :+1:

config/application.rb
LogaJ: Agreed, I'll check with Steve.

andistuder: When considering the default per page results, it's less about the DB, but about instantiating `n` ActiveRecord objects and rendering the view. Whatever Steve settles to, let's review load time once we have it deployed.

PR #151, opened 2016-03-23
Remove environment from Heroku DeployEvent app name

dideler:
In our latest snapshot repository, `ReleasedTicketRepository`, we load Git repositories via the app name that we get from DeployEvents. This would fail for Heroku DeployEvents, because most Heroku apps follow a naming convention of `APPNAME-ENVIRONMENT`, e.g. "awesomeapp-staging", but the repository name would be "awesomeapp".

Also includes refactoring of `DeployEvent` model. Previously we handled Heroku payloads via fallbacks  but that behaviour was difficult to follow. It's now more explicit how we handle the payload depending on if it's a Heroku deploy. The plan was to then extract it to its own class using STI but it's a difficult change because it would require a new `EventType` and endpoint, and changing the webhook endpoint for all onboarded Heroku apps is not trivial. So that's deferred.

### DIFF
--- a/app/models/events/deploy_event.rb
+++ b/app/models/events/deploy_event.rb
@@ -6,7 +6,11 @@ module Events
     ENVIRONMENTS = %w(uat staging production).freeze
 
     def app_name
-      (details['app_name'] || details['app'])&.downcase
+      if deployed_to_heroku?
+        details['app']&.downcase
+      else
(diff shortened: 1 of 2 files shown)


andistuder:
LGTM

LogaJ:
:+1:

PR #154, opened 2016-03-23
Optimize ticket snapshotting

dideler:
Only snapshot tickets if they contain a Feature Review.

##### Context
JIRA can be configured to send webhook events for all Issue activity, which results in a massive amount of ticket snapshots -- one for every ticket.

There is no need to prematurely snapshot tickets without a Feature Review because many of them will never contain a Feature Review and hence do not go through the Shipment Tracker process.

This will not cause any issues when we *do* have to snapshot tickets because the payload of comment events (which is what ST receives when a FR is linked) contain the all the other information we need to create the snapshots, such as ticket key, title (summary), description, etc.

---

- [x] Only snapshot tickets if they contain a FR or a previous snapshot exists for the ticket
- [x] Only record `approved_at` time on ticket snapshots for approval events or if time can be carried over
  - Needed for "requires re-approval" scenario to continue working with snapshotting optimization
- [ ] ~~Use JIRA timestamp instead of record creation time for `approved_at`~~  
  Excluded from this PR and moved to #168

### DIFF
--- a/app/jobs/relink_ticket_job.rb
+++ b/app/jobs/relink_ticket_job.rb
@@ -12,11 +12,11 @@ class RelinkTicketJob < ActiveJob::Base
     full_repo_name = args.delete(:full_repo_name)
     branch_created = args.delete(:branch_created)
 
-    linked_tickets = relink_tickets(before_sha, after_sha) unless branch_created
-
-    post_not_found_status(full_repo_name: full_repo_name, sha: after_sha) \
-      if branch_created || linked_tickets&.empty?
-    post_error_status(full_repo_name: full_repo_name, sha: after_sha) if @send_error_status
+    if branch_created || relink_tickets(before_sha, after_sha).empty?
+      post_not_found_status(full_repo_name: full_repo_name, sha: after_sha)
+    elsif @send_error_status
+      post_error_status(full_repo_name: full_repo_name, sha: after_sha)
+    end
   end
 
   private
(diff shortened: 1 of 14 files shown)


0mega:
LGTM

db/migrate/20160324142505_add_defaults_to_array_columns.rb
```diff
@@ -0,0 +1,9 @@
+class AddDefaultsToArrayColumns < ActiveRecord::Migration
+  def change
```
andistuder: pls test if rollback possible

app/jobs/relink_ticket_job.rb
```diff
@@ -12,11 +12,11 @@ def perform(args)
...
     branch_created = args.delete(:branch_created)
 
-    linked_tickets = relink_tickets(before_sha, after_sha) unless branch_created
-
-    post_not_found_status(full_repo_name: full_repo_name, sha: after_sha) \
-      if branch_created || linked_tickets&.empty?
-    post_error_status(full_repo_name: full_repo_name, sha: after_sha) if @send_error_status
+    if branch_created || relink_tickets(before_sha, after_sha).empty?
+      post_not_found_status(full_repo_name: full_repo_name, sha: after_sha)
+    elsif @send_error_status
```
andistuder: pls check if this conditional is still working the same as before. ie. fall back to `@send_error_status`

andistuder: Resolved: I've manually tested this - it still works as expected. Also, there is sufficient test coverage for that case.

0mega:
👍

andistuder:
Scheduled for deployment Tuesday, 12 April 2016, 8AM BST

PR #158, opened 2016-04-01
fix deploy alert errors

andistuder:
ensure that deploy alerts are raised even if the deploy events contain SHAs that can not be parsed or looked up.


andistuder:
ping @FundingCircle/black-bat

LogaJ:
LGTM

PR #159, opened 2016-04-04
Improve Dashboard UI

0mega:
- [x] Moves Dashboard to home page and moves home page contents to wiki
- [x] Improves ticket summary style for cases when words don't fit in one line
- [x] Makes search text optional
- [x] Fix 'result' text to reflect user's search criteria
- [x] Improve search input placeholder
- [x] Show deploys for current day by default


0mega:
ping @FundingCircle/black-bat

andistuder:
LGTM

dideler:
When I spoke with @steveweet he mentioned showing deploys for current day as default (without user query). Is that change going to be part of this PR?

app/views/home/index.html.haml
```diff
@@ -1,85 +1,39 @@
...
-  A deployer can
-  =link_to "view pending and past deploys to production", releases_url
-  for a specific app. We call these "releases".
-%p
-  A release should have a review associated with it. A link to the review page
-  along with its current status appears in the right-most column.
-%p
-  A release will be flagged as red if it does not have a review, or it does but the work has not been
-  approved (i.e. status is not "Done").
+- if @query || @from_date || @to_date
```
dideler: When I search without any input, all results are returned (limited to first 100) due to these being empty strings. We should tighten this check to make sure they are not empty.

app/controllers/home_controller.rb
```diff
@@ -3,18 +3,15 @@ class HomeController < ApplicationController
...
     @query = params[:q]
     @from_date = params[:from]
     @to_date = params[:to]
 
-    @tickets = if @query
-                 released_ticket_repo.tickets_for_query(build_query_hash(@query, params))
-               else
-                 []
-               end
+    redirect_to root_path(from: Time.zone.today, to: Time.zone.today) if @query.blank? &&
```
dideler: **Not a blocker:** can extract the condition to a private method, e.g. `blank_search?`

app/controllers/home_controller.rb
```diff
@@ -3,18 +3,15 @@ class HomeController < ApplicationController
...
+                                                                         @from_date.blank? &&
+                                                                         @to_date.blank?
 
-    render 'dashboard'
+    @tickets = released_ticket_repo.tickets_for_query(build_query_hash(@query, params))
```
dideler: **Not a blocker:** Can just pass params and extract query from it

dideler:
LGTM

PR #160, opened 2016-04-04
Improve parsing of JIRA comments

dideler:
When a Feature Review URL in a JIRA comment is contained within parentheses such as `(url)`, then we do not strip out the closing parenthesis because it's a valid URI. But this causes issues when we try and post a GitHub Commit Status for example, since the SHA includes a `)`.

### DIFF
--- a/app/models/factories/feature_review_factory.rb
+++ b/app/models/factories/feature_review_factory.rb
@@ -51,7 +51,8 @@ module Factories
     end
 
     def clean_uri(uri)
-      uri.chomp(']') # when parsing text with Jira markup we sometimes find a trailing ']'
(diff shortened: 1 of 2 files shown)


0mega:
I couldn't find a test using ')' at the end of ST link. Should we add that as a regression test?

dideler:
@0mega I had a separate unit test for it at first but then combined it with the test that has `]).` at the end because I didn't want separate tests for each case where there is a trailing non-word character.

andistuder:
LGTM

0mega:
@dideler that's fine, I didn't see it was included in on this line https://github.com/FundingCircle/shipment_tracker/pull/160/files#diff-09b781cf7c1296095b6274a806daebceR49

PR #161, opened 2016-04-04
Remove Feature Review search

dideler:
The Feature Review search was an unpolished and underused feature. It's now made obsolete by the dashboard search which also allows you to find FRs (via their JIRA tickets) given a commit SHA. The dashboard search is also easier to use since it doesn't require you to select the application that the SHA belongs to.

- [x] Remove "Search" from navbar
- [x] Remove route
- [x] Remove controller action
- [x] Remove views/pages
- [x] Remove any related dead code
- [x] Change GH Commit Status link to use dashboard search instead of FR search


dideler:
@FundingCircle/black-bat ready for review.

I've removed any related dead code I could easily find. I'd like to use code coverage to find more dead code but I think there's an issue with our current coverage since the stats it's reporting are unusual. Going to try updating the gem and see if that fixes it, but if not, not a blocker.

app/controllers/feature_reviews_controller.rb
```diff
@@ -58,6 +39,10 @@ def link_ticket
 
   private
 
+  def ticket_linking_options
```
andistuder: nice

PR #164, opened 2016-04-06
Handle exception when failing to expand a deploy version

dideler:
An uncaught exception is raised when we have DeployEvents for an app but the git repository is missing, because it was never onboarded or it was removed. This rescues the exception (so the updater doesn't blow up) and notifies Honeybadger until we have a better solution.

This was happening whenever snapshotting DeployEvents, which happens in a couple of Repositories.

@FundingCircle/black-bat please review -- this issue is currently present on staging and development (if you use staging data), and is bound to happen on production.

### DIFF
--- a/app/models/deploy_alert.rb
+++ b/app/models/deploy_alert.rb
@@ -34,7 +34,7 @@ class DeployAlert
   end
 
   def self.alert_unknown_version(deploy)
-    "#{alert_header(deploy)}Deploy event sent to Shipment Tracker is missing the software version."
+    "#{alert_header(deploy)}Deploy event sent to Shipment Tracker contains an unknown software version."
   end
 
   def self.alert_rollback(deploy)
--- a/app/models/events/deploy_event.rb
+++ b/app/models/events/deploy_event.rb
@@ -85,7 +85,9 @@ module Events
     end
 
     def full_sha(sha)
-      return sha if sha.nil? || sha.length == 40
+      return if sha.nil? || sha.length > 40
+      return sha if sha.length == 40
+
       git_repository_loader = GitRepositoryLoader.from_rails_config
       repo = git_repository_loader.load(app_name)
       repo.commit_for_version(sha).id
--- a/app/models/events/deploy_event.rb
+++ b/app/models/events/deploy_event.rb
@@ -26,7 +26,7 @@ module Events
       @version ||= if deployed_to_heroku?
                      details['head_long']
                    else
-                     expand_sha(details['version'])
+                     full_sha(details['version'])
                    end
     end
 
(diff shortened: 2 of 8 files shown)


app/models/repositories/released_ticket_repository.rb
```diff
@@ -81,15 +81,13 @@ def snapshot_jira_event(event)
...
-        end
-      rescue GitRepositoryLoader::NotFound, GitRepository::CommitNotValid, GitRepository::CommitNotFound => e
-        log_warning(e, event)
+
+      last_deploy = latest_production_deploy(event.app_name, event.locale, event.created_at)
+      released_commits(event.app_name, event.version, last_deploy&.version).each do |commit|
+        update_ticket_deploy_data(event, commit)
       end
+    rescue GitRepositoryLoader::NotFound, GitRepository::CommitNotValid, GitRepository::CommitNotFound => e
+      Rails.logger.warn "Could not snapshot DeployEvent #{event.id} for '#{event.app_name}'\n#{e.message}"
```
dideler: Not sure if we should also be notifying Honeybadger here?

0mega: Sending this to Honeybadger will build a "backlog" of problematic apps, showing which send wrong/strange deploy events. But it will also include some false positives.
Ultimately I think we want to see this on HB, so that we can correct the wrong setups.

andistuder: We discussed this offline: we don't report to honeybadger here as it's covered already when we process the same event in the Deploy Repo. There it's more critical we run a deploy_alert also.

app/models/deploy_alert.rb
```diff
@@ -34,7 +34,7 @@ def self.alert_not_on_master(deploy)
   end
 
   def self.alert_unknown_version(deploy)
-    "#{alert_header(deploy)}Deploy event sent to Shipment Tracker is missing the software version."
+    "#{alert_header(deploy)}Deploy event sent to Shipment Tracker contains an unknown software version."
```
andistuder: Great! thanks

dideler:
@FundingCircle/black-bat [added one more commit](e566f40182d42a8a83983ccd3104deeb1b0acb42) to handle cases where version is longer than 40 chars.

app/models/events/deploy_event.rb
```diff
@@ -84,8 +84,8 @@ def heroku_app_name_prefix
       heroku_app_name.split('-').first.downcase
     end
 
-    def expand_sha(sha)
-      return sha unless sha.present? && sha.size < 40
+    def full_sha(sha)
+      return sha if sha.nil? || sha.length == 40
```
dideler: Cheaper if we handle invalid SHAs with a guard clause instead of rescuing exceptions up the stack.
```ruby
return if sha.nil? || sha.length > 40
return sha if sha.length == 40
```

0mega:
LGTM

PR #165, opened 2016-04-06
remove events key from payload

andistuder:
this seems added as a duplication by the rails controller

and we do not use it

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -3,7 +3,7 @@ class EventsController < ApplicationController
   skip_before_action :verify_authenticity_token
 
   def create
-    event_factory.create(params[:type], request.request_parameters, current_user.email)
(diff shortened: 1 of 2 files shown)


app/controllers/events_controller.rb
```diff
@@ -23,4 +23,10 @@ def unauthenticated_strategy
     self.status = 403
     self.response_body = 'Forbidden'
   end
+
+  def event_payload(parameters)
+    event_payload = parameters.dup
+    event_payload.delete 'event'
+    event_payload
```
dideler: Can reduce the code with [`Hash#except` from Rails](http://api.rubyonrails.org/classes/Hash.html#method-i-except)

dideler:
Interesting, maybe some side effect from Rails STI?

app/controllers/events_controller.rb
andistuder: thanks! good idea

dideler:
LGTM

PR #166, opened 2016-04-07
Handle invalid dates for dashboard search

dideler:
A flash warning will now be shown if an invalid date is entered.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -7,26 +7,26 @@ class HomeController < ApplicationController
     @from_date = params[:from]
     @to_date = params[:to]
 
-    redirect_to root_path(from: Time.zone.today, to: Time.zone.today) if @query.blank? &&
-                                                                         @from_date.blank? &&
-                                                                         @to_date.blank?
+    search_for_tickets_deployed_today if empty_search?
 
-    @tickets = released_ticket_repo.tickets_for_query(build_query_hash(@query, params))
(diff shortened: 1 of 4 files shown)


app/controllers/home_controller.rb
```diff
@@ -7,26 +7,29 @@ def index
     @from_date = params[:from]
     @to_date = params[:to]
 
-    redirect_to root_path(from: Time.zone.today, to: Time.zone.today) if @query.blank? &&
-                                                                         @from_date.blank? &&
-                                                                         @to_date.blank?
+    search_for_tickets_deployed_today if empty_search?
```
0mega: nice refactor

app/views/home/index.html.haml
```diff
@@ -10,30 +10,28 @@
       %input#to_date.input-sm.form-control{:name => "to", :type => "text", :style => "width: 120px; height: 34px;", :value => @to_date}/
   %input.btn.btn-primary{:type => "submit", :value => "Search"}/
 
-- if @query || @from_date || @to_date
```
dideler: Removed this because now a date will always be set, so the condition didn't matter.

0mega:
LGTM

app/controllers/home_controller.rb
andistuder: No need to change, but for future reference: I'd prefer if the private method name is explicit the method it calls on ActionController.

PR #167, opened 2016-04-07
Include timezone in releases deployed date

0mega:
So that it's consistent with dashboard page

### DIFF
--- a/app/helpers/releases_helper.rb
+++ b/app/helpers/releases_helper.rb
@@ -12,4 +12,8 @@ module ReleasesHelper
       link_to(feature_review.authorisation_status.to_s.humanize, feature_review.path)
     end
   end
+
+  def time_with_timezone(time)
+    time&.strftime('%F %H:%M %Z')
+  end
 end
--- a/app/views/releases/show.html.haml
+++ b/app/views/releases/show.html.haml
@@ -44,4 +44,4 @@
       %td
         - release.feature_reviews.each do |feature_review|
           %div= feature_review_link(feature_review)
-      %td= release.production_deploy_time.try(:to_formatted_s, :long_ordinal)
+      %td= time_with_timezone(release.production_deploy_time)


app/helpers/releases_helper.rb
```diff
@@ -12,4 +12,8 @@ def feature_review_link(feature_review)
       link_to(feature_review.authorisation_status.to_s.humanize, feature_review.path)
     end
   end
+
+  def long_ordinal_with_timezone(time)
+    time.try(:strftime, "%B #{time.try(:day).try(:ordinalize)}, %Y %H:%M %Z")
```
dideler: For readability, could we instead do
```ruby
time.strftime("%B #{time.day.ordinalize}, %Y %H:%M %Z") if time
# or
time&.strftime("%B #{time&.day&.ordinalize}, %Y %H:%M %Z")
```

0mega: sure, I prefer the second one :), will change

dideler: Since we're adding more text to the column, I'm in favour of removing the ordinalize (which turns 7 into 7th for example). Saving a few characters could prevent most lines being split into two rows for making the content fit.

dideler:
👍

One of my comments got buried during rebase, so reposting here.

> Since we're adding more text to the column, I'm in favour of removing the ordinalize (e.g. 7 => 7th). to save a few characters. This could prevent many lines being split into two rows which hurts presentation. But this a product decision /cc @steveweet

0mega:
@FundingCircle/black-bat

dideler:
LGTM

andistuder:
LGTM - for discussion: does it add value if we dry this up: 
https://github.com/FundingCircle/shipment_tracker/blob/master/app/models/repositories/released_ticket_repository.rb#L111 

from
`deployed_at: event.created_at.strftime('%F %H:%M %Z'),`
to 
`deployed_at: ActionController::Base.helpers.time_with_timezone(event.created_at),`

dideler:
I'd keep it as is so we don't start using helpers in models. If we used a decorator instead then 👍

PR #169, opened 2016-04-08
Improve Dashboard

0mega:
- Improves search result messages
- Moves ticket searching logic from controller to a use case
- Replaces home controller content by Dashboard


app/controllers/home_controller.rb
```diff
@@ -3,33 +3,14 @@ class HomeController < ApplicationController
   SHA_REGEX = /\b[0-9a-f]{40}\b/
```
andistuder: redundant

PR #171, opened 2016-05-23
fix deploy alerts for dependent releases

andistuder:
Change to initialize `Queries::ReleasesQuery` with commits correct order: 'newest-first'. This is the order that the query uses if no commits are passed in - see: http://github.com/FundingCircle/shipment_tracker/blob/3e5a343af249a143c770b261c16a274c163ccc1c/app/models/queries/releases_query.rb#L39-L40

This fixes a bug where the deploy auditor does not detect unauthorised releases that are deployed as bundled of a newer release. 

paired with @0mega

### DIFF
--- a/app/models/deploy_alert.rb
+++ b/app/models/deploy_alert.rb
@@ -92,8 +92,11 @@ class DeployAlert
     def auditable_commits_newest_first
       @commits ||= if previous_deploy
                      git_repo
-                       .commits_between(previous_deploy.version, current_deploy.version, simplify: true)
-                       .reverse
+                       .commits_between(previous_deploy.version,
+                         current_deploy.version,
+                         simplify: true,
+                         newest_first: true,
+                                       )
                    else
                      [git_repo.commit_for_version(current_deploy.version)]
                    end
--- a/app/models/deploy_alert.rb
+++ b/app/models/deploy_alert.rb
@@ -69,7 +69,9 @@ class DeployAlert
     end
 
     def recent_releases_authorised?
-      release_query = release_query_for(auditable_commits, current_deploy.region, current_deploy.app_name)
+      release_query = release_query_for(auditable_commits_newest_first,
+        current_deploy.region,
+        current_deploy.app_name)
       release_query.deployed_releases.all?(&:authorised?)
     end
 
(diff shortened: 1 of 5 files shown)


andistuder:
ping @dideler @steveweet @LogaJ

app/models/deploy_alert.rb
```diff
@@ -87,9 +89,11 @@ def release_query_for(auditable_commits, region, app_name)
       )
     end
 
-    def auditable_commits
+    def auditable_commits_newest_first
       @commits ||= if previous_deploy
-                     git_repo.commits_between(previous_deploy.version, current_deploy.version, simplify: true)
+                     git_repo
+                       .commits_between(previous_deploy.version, current_deploy.version, simplify: true)
+                       .reverse
```
dideler: Would be nicer if we could override [default walker sorting behaviour](https://github.com/FundingCircle/shipment_tracker/blob/857550bf51d232dbd62ad4fb15ed6ede7b2aa744/app/models/git_repository.rb#L140) to save some computation. Not a blocker.

dideler:
LGTM - pending green build

app/models/deploy_alert.rb
andistuder: thanks for feedback... I had a bit more wind in my sails today, and refactored as suggested!.

0mega:
LGTM 👍 good catch

andistuder:
Cheers for @LS80 for reporting this!

spec/models/git_repository_spec.rb
```diff
@@ -120,6 +120,15 @@
         }.to raise_error(GitRepository::CommitNotFound, non_existent_commit)
       end
     end
+
+    context 'when called with new_first flag set true' do
```
dideler: s/new_first/newest_first/

spec/models/git_repository_spec.rb
```diff
@@ -120,6 +120,15 @@
...
+      let(:options) { { newest_first: true } }
+
+      it 'returns all commits between two commits, including the end commit' do
```
dideler: Suggest something like

> returns all commits between two commits in order of newest first, ...

to differentiate from the example on L90

dideler:
Thanks, LGTM 👍

spec/models/git_repository_spec.rb
andistuder: oops.. sorry, merged too quickly.

andistuder: See follow-up PR https://github.com/FundingCircle/shipment_tracker/pull/173

spec/models/git_repository_spec.rb
andistuder: as above: https://github.com/FundingCircle/shipment_tracker/pull/173

PR #175, opened 2016-06-06
Add more context to error logging

dideler:
/cc @0mega

### DIFF
--- a/lib/tasks/jobs.rake
+++ b/lib/tasks/jobs.rake
@@ -88,9 +88,7 @@ namespace :jobs do
             begin
               loader.load(app_name, update_repo: true)
             rescue StandardError => error
-              Honeybadger.context(app_name: app_name, remote_head: repos_hash_changed[app_name])
-              Honeybadger.notify(error)
-              Honeybadger.context.clear!
+              Honeybadger.notify(error, context: { app_name: app_name, remote_head: repos_hash_changed[app_name] })
             end
           end
         end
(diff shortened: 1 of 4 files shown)


0mega:
👍

lib/tasks/jobs.rake
```diff
@@ -83,7 +85,13 @@ namespace :jobs do
...
           group.compact.each do |app_name|
             break if @shutdown
-            loader.load(app_name, update_repo: true)
+            begin
+              loader.load(app_name, update_repo: true)
+            rescue StandardError => error
+              Honeybadger.context(app_name: app_name, remote_head: repos_hash_changed[app_name])
+              Honeybadger.notify(error)
+              Honeybadger.context.clear!
+            end
```
damien: FWIW, if you want a one-time Honeybadger context, the `notify` method will accept a context via a keyword argument:

```ruby
Honeybadger.notify(error, context: { app_name: app_name, remote_head: repos_hash_changed[app_name] })
```

This will remove any risk of having the context setting/unsetting get out of sync. Not sure how HB works in a threaded environment, but this may also be safer from a concurrency standpoint as well.

dideler: Neat, thanks!

dideler: 0062a0ea37a7169ee895f0e769e997e86c83942d

dideler:
Tested the inline context, works: https://app.honeybadger.io/projects/44071/faults/30135192#notice-trace

PR #183, opened 2016-08-15
Fix perfomance/projections_spec.rb feature_review url

your:
In this block: 

```ruby
..
  let(:url) { feature_review_path(apps, server) }
  let(:path) { URI.parse(url).request_uri }
..
```

`request_uri` cannot be called on a URI parsed object without a real URL -- `feature_review_path` is stripping it down so you have to use `feature_review_url` first, or `path` on the URI parsed object

Note: this was always passing because CircleCI is configured to skip this test, it would fail otherwise

### DIFF
--- a/spec/performance/projections_spec.rb
+++ b/spec/performance/projections_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe 'Projection performance', type: :request do
   describe 'Feature Review' do
     let(:apps) { { 'frontend' => 'abc', 'backend' => 'def' } }
     let(:server) { 'uat.fc.com' }
-    let(:url) { feature_review_path(apps, server) }
+    let(:url) { feature_review_url(apps, server) }
     let(:path) { URI.parse(url).request_uri }
 
     it 'measures the request time' do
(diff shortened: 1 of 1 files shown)


dideler:
Some background on performance specs. We created them to benchmark how long it took to apply events for certain page requests, and then justify to the business that we needed to spend time on optimizing this by using snapshots to avoid re-applying events from the beginning every time.

After that we stopped maintaining them, so they're probably very outdated. For example, "Feature Review Search" doesn't exist anymore (L40).

your:
Thank you for the explanation @dideler !

If I run `bundle exec rspec` I'm expecting to have all of them working. Not a priority of course, but they should be either fixed or definitely removed.  What do you think?

dideler:
Good observation @your! Something that we forgot to document in the README is the suggested way to run tests. We used `bundle exec rake` which will run rspec (unit) tests, cucumber (acceptance) tests, and rubocop (linter) - in that order. This excludes performance tests.

See https://github.com/FundingCircle/shipment_tracker/blob/master/Rakefile#L11-L17

We should definitely mention this in the README. As for the fixing/removing the perf tests, I'm okay with removing them. They mostly focus on snapshotting, which is already done. There are other areas that can be optimized, but they may not be properly captured by these tests so the tests aren't adding much value (at least in their current state).

FraDim:
LGTM

PR #186, opened 2016-09-01
Ensure get_dependent_commits returns an empty array when commit is not found

your:
Rescue `CommitNotFound` and return empty array when calling `get_dependent_commits`

### DIFF
--- a/app/decorators/feature_review_with_statuses.rb
+++ b/app/decorators/feature_review_with_statuses.rb
@@ -31,6 +31,8 @@ class FeatureReviewWithStatuses < SimpleDelegator
     app_versions.map do |app_name, version|
       latest_commit = fetch_commit_for(app_name, version)
 
+      latest_commit = GitCommit.new(id: version) unless latest_commit.id
+
       [app_name, latest_commit]
     end
   end
--- a/app/decorators/feature_review_with_statuses.rb
+++ b/app/decorators/feature_review_with_statuses.rb
@@ -28,11 +28,11 @@ class FeatureReviewWithStatuses < SimpleDelegator
   end
 
   def apps_with_latest_commit
-    app_versions.map do |app_name, version|
+    app_versions.map { |app_name, version|
       latest_commit = fetch_commit_for(app_name, version)
 
-      [app_name, latest_commit]
-    end
+      latest_commit ? [app_name, latest_commit] : nil
+    }.compact
   end
 
   def build_status
--- a/app/decorators/feature_review_with_statuses.rb
+++ b/app/decorators/feature_review_with_statuses.rb
@@ -28,11 +28,11 @@ class FeatureReviewWithStatuses < SimpleDelegator
   end
 
   def apps_with_latest_commit
-    app_versions.map { |app_name, version|
+    app_versions.map do |app_name, version|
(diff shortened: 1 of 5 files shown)


FraDim:
LGTM

app/models/git_repository.rb
```diff
@@ -71,7 +71,7 @@ def get_dependent_commits(commit_oid)
     end
 
     dependent_commits + commits_between(common_ancestor_oid, commit_oid)[0...-1]
-  rescue CommitNotValid
+  rescue CommitNotValid, CommitNotFound
     []
```
berkos: 💭  does it worth to log something when we rescue?

0mega:
Only concern is how this part will look, you might still get an exception when trying to extract the commit details from a nil commit: https://github.com/FundingCircle/shipment_tracker/blob/4e526452c1b6b1ca4f5cc9eb2a3995b1c5181625/app/views/feature_reviews/show.html.haml#L55-L57.

Did you test it on heroku to see what happens when the commit is not found?

ioannaK89:
@0mega You are right! It actually fails there. We are working on fixing it

app/decorators/feature_review_with_statuses.rb
```diff
@@ -28,11 +28,11 @@ def github_repo_urls
...
 
   def apps_with_latest_commit
-    app_versions.map do |app_name, version|
+    app_versions.map { |app_name, version|
       latest_commit = fetch_commit_for(app_name, version)
 
-      [app_name, latest_commit]
-    end
+      latest_commit ? [app_name, latest_commit] : nil
+    }.compact
```
0mega: In this case, if a Feature Review(FR) has app1 and app2 associated, but at some point the commit for app1 could not be found, it will look like the FR is only linked to app2. It might be better to build manually a commit with hardcoded values so that users are not mislead.
You could use http://github.com/FundingCircle/shipment_tracker/blob/4e526452c1b6b1ca4f5cc9eb2a3995b1c5181625/app/models/git_repository.rb#L159-L160 to create a commit with "Unknown" user and "Unknown" time and the original commit id for the app1 from the query params.

app/models/git_repository.rb
```diff
@@ -165,6 +166,13 @@ def build_commit(commit)
     )
   end
 
+  def build_unresolved_commit(sha)
+    GitCommit.new(
+      id: sha,
+      author_name: 'unknown',
```
Antti: Remove `author_name` from here.

app/models/git_repository.rb
```diff
@@ -46,7 +46,8 @@ def recent_commits_on_main_branch(count = 50)
     build_commits(walker.take(count))
   end
 
-  def commit_for_version(sha)
+  def commit_for_version(sha, lookup: true)
+    return build_unresolved_commit(sha) unless lookup
     build_commit(lookup(sha))
```
Antti:
```
commit = build_commit(lookup(sha))
commit.id = sha unless commit.id
```

app/models/git_repository.rb
your: I don't think this is critical enough, at least for now..

0mega:
LGTM 👍

PR #191, opened 2016-11-14
Update outdated gems

sgerrand:
💁  These changes updates all of the outdated gems which don't have constraints.


Gemfile
```diff
@@ -12,7 +12,7 @@ gem 'git_clone_url'
 gem 'haml-rails'
 gem 'has_secure_token'
 gem 'honeybadger', '~> 2.0'
-gem 'jira-ruby', require: 'jira'
+gem 'jira-ruby'
```
damien: Is this correct? `require` isn't a constraint, it's the reference Bundler uses when it generates a `require` statement. Without this Bundler will try to run `require 'jira-ruby'` instead of `require 'jira'`.

Gemfile.lock
```diff
@@ -213,68 +219,69 @@ GEM
...
+    rspec-rails (3.5.2)
+      actionpack (>= 3.0)
+      activesupport (>= 3.0)
+      railties (>= 3.0)
+      rspec-core (~> 3.5.0)
+      rspec-expectations (~> 3.5.0)
+      rspec-mocks (~> 3.5.0)
+      rspec-support (~> 3.5.0)
+    rspec-support (3.5.0)
+    rubocop (0.45.0)
```
damien: This may be the cause of some of your test failures. I've noticed in the past that Rubocop is pretty active and will add/remove/change rules and rule configs with each version bump.

Unless you're ready to update/refresh your rubocop rules, I'd suggest pinning this to the version your current rules are configured for.

Gemfile
sgerrand: Yes. The `jira-ruby` gem changed in v1.0. That `jira` file no longer exists in the gem.

Gemfile.lock
sgerrand: Not fussed either way, I'll follow your advice! :family_man_boy:

sgerrand: @damien: Updated in 92409d4a4e13051d5a091db1c192579b94a6bb5a. ♻️

sgerrand:
Superseded by ⏳

PR #192, opened 2016-11-23
NS-1506: Add ability to unlink Jira ticket from a feature review

MiroslavCsonka:
### What does this PR do?
Adds ability to unlink a Jira ticket from a feature review

### Where should the reviewer start?
Posting a comment to Jira - `app/use_cases/unlink_ticket.rb`
Processing webhook Jira notification - `app/models/repositories/ticket_repository.rb`

### Any background context you want to provide?
*intentionally left blank*

### Any specific implementation changes that would benefit highlighting?
I've decided to use the same process as we use for linking tickets (ST -> Jira -> ST)

### What are the relevant tickets / PRs?
- [x] JIRA :gem:: https://jira.fundingcircle.com/browse/NS-1506

### DIFF
--- a/app/models/events/handlers/ticket_handler.rb
+++ b/app/models/events/handlers/ticket_handler.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+module Events
+  module Handlers
+    class TicketHandler
+      def initialize(ticket, event)
+        @ticket = ticket
+        @event = event
+      end
+
+      def apply
+        ticket.merge(
+          'key' => event.key,
+          'summary' => event.summary,
+          'status' => event.status,
+          'event_created_at' => event.created_at,
+          'approved_at' => merge_approved_at(ticket, event),
+        )
+      end
+
+      protected
+
+      attr_reader :ticket, :event
+
+      def feature_reviews_from_event
+        Factories::FeatureReviewFactory.new.create_from_text(event.comment)
+      end
+
+      private
+
+      def merge_approved_at(last_ticket, event)
+        if event.approval?
+          event.created_at
+        elsif last_ticket.present? && Ticket.new(status: event.status).approved?
+          last_ticket['approved_at']
+        end
+      end
+    end
+  end
+end
--- a/app/models/repositories/ticket_repository.rb
+++ b/app/models/repositories/ticket_repository.rb
@@ -20,17 +20,16 @@ module Repositories
     def tickets_for_path(feature_review_path, at: nil)
       query = at ? store.arel_table['event_created_at'].lteq(at) : nil
       store
-        .select('DISTINCT ON (key) *')
         .where('paths @> ARRAY[?]', feature_review_path)
-        .where(query)
+        .where(id: store.select('MAX(id) as id').where(query).group(:key))
         .order('key, id DESC')
         .map { |t| Ticket.new(t.attributes) }
     end
 
     def tickets_for_versions(versions)
       store
-        .select('DISTINCT ON (key) *')
         .where('versions && ARRAY[?]', versions)
+        .where(id: store.select('MAX(id) as id').group(:key))
         .order('key, id DESC')
         .map { |t| Ticket.new(t.attributes) }
     end
@@ -43,10 +42,11 @@ module Repositories
 
       return if last_ticket.empty? && feature_reviews.empty?
 
-      new_ticket = build_ticket(last_ticket, event, feature_reviews)
+      new_ticket = event.apply(last_ticket)
+
       store.create!(new_ticket)
 
-      update_github_status_for(new_ticket) if update_github_status?(event, feature_reviews)
+      update_github_status_for([new_ticket, last_ticket]) if update_github_status?(event, feature_reviews)
     end
 
     private
@@ -63,50 +63,10 @@ module Repositories
       event.approval? || event.unapproval? || feature_reviews.present?
     end
 
-    def build_ticket(last_ticket, event, feature_reviews)
-      last_ticket.merge(
-        'key' => event.key,
-        'summary' => event.summary,
-        'status' => event.status,
-        'paths' => merge_ticket_paths(last_ticket, feature_reviews),
-        'event_created_at' => event.created_at,
-        'versions' => merge_ticket_versions(last_ticket, feature_reviews),
-        'approved_at' => merge_approved_at(last_ticket, event),
-        'version_timestamps' => merge_version_timestamps(last_ticket, feature_reviews, event),
-      )
-    end
+    def update_github_status_for(ticket_hashes)
+      tickets = ticket_hashes.map { |ticket| Ticket.new(ticket) }
 
-    def merge_version_timestamps(ticket, feature_reviews, event)
-      old_version_timestamps = ticket.fetch('version_timestamps', {})
-      new_version_timestamps = feature_reviews.flat_map(&:versions).each_with_object({}) { |version, hash|
-        hash[version] = event.created_at
-      }
-      new_version_timestamps.merge!(old_version_timestamps)
-    end
-
-    def merge_ticket_paths(ticket, feature_reviews)
-      old_paths = ticket.fetch('paths', [])
-      new_paths = feature_reviews.map(&:path)
-      old_paths.concat(new_paths).uniq
-    end
-
-    def merge_ticket_versions(ticket, feature_reviews)
-      old_versions = ticket.fetch('versions', [])
-      new_versions = feature_reviews.flat_map(&:versions)
-      old_versions.concat(new_versions).uniq
-    end
-
-    def merge_approved_at(last_ticket, event)
-      if event.approval?
-        event.created_at
-      elsif last_ticket.present? && Ticket.new(status: event.status).approved?
-        last_ticket['approved_at']
-      end
-    end
-
-    def update_github_status_for(ticket_hash)
-      ticket = Ticket.new(ticket_hash)
-      array_of_app_versions = feature_review_factory.create_from_tickets([ticket]).map(&:app_versions)
+      array_of_app_versions = feature_review_factory.create_from_tickets(tickets).map(&:app_versions)
 
       array_of_app_versions.map(&:invert).reduce({}, :merge).each do |version, app_name|
         repository_location = git_repository_location.find_by_name(app_name)
--- a/app/use_cases/link_ticket.rb
+++ b/app/use_cases/link_ticket.rb
@@ -4,25 +4,30 @@ require 'clients/jira'
 
 class LinkTicket
   include SolidUseCase
+  include TicketValidationHelper
+
+  COMMENT_LABEL = 'Feature ready for review'
 
   steps :validate_id_format, :assert_not_linked, :post_comment
 
-  def validate_id_format(args)
-    return fail :invalid_key, message: invalid_key_message(args) unless /[A-Z][A-Z]+-\d*/ =~ args[:jira_key]
-    continue(args)
+  class << self
+    def build_comment(url)
+      "[#{COMMENT_LABEL}|#{url}]"
+    end
   end
 
   def assert_not_linked(args)
     tickets = Repositories::TicketRepository.new.tickets_for_path(args[:feature_review_path])
 
-    return fail :duplicate_key, message: duplicate_key_message(args) if tickets.any? { |ticket|
-      ticket.key == args[:jira_key]
-    }
+    has_links = tickets.any? { |ticket| ticket.key == args[:jira_key] }
+    return fail :duplicate_key, message: duplicate_key_message(args) if has_links
     continue(args)
   end
 
   def post_comment(args)
-    JiraClient.post_comment(args[:jira_key], jira_comment(args))
+    feature_review_url = "#{args[:root_url].chomp('/')}#{args[:feature_review_path]}"
+
+    JiraClient.post_comment(args[:jira_key], self.class.build_comment(feature_review_url))
     continue(success_message(args))
   rescue JiraClient::InvalidKeyError
     fail :invalid_key, message: invalid_key_message(args)
@@ -46,13 +51,6 @@ class LinkTicket
   end
 
   def success_message(args)
-    "Feature Review was linked to #{args[:jira_key]}."\
-    ' Refresh this page in a moment and the ticket will appear.'
-  end
-
-  def jira_comment(args)
-    feature_review_url = "#{args[:root_url].chomp('/')}#{args[:feature_review_path]}"
-
-    "[Feature ready for review|#{feature_review_url}]"
+    "Feature Review was linked to #{args[:jira_key]}. Refresh this page in a moment and the ticket will appear."
   end
 end
--- a/app/use_cases/unlink_ticket.rb
+++ b/app/use_cases/unlink_ticket.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+require 'clients/jira'
+
+class UnlinkTicket
+  COMMENT_LABEL = 'Feature review was unlinked'
+
+  include SolidUseCase
+  include TicketValidationHelper
+
+  steps :validate_id_format,
+    :assert_linked,
+    :post_unlink_comment
+
+  class << self
+    def build_comment(feature_review_url)
+      "[#{COMMENT_LABEL}|#{feature_review_url}]"
+    end
+  end
+
+  def assert_linked(args)
+    tickets = Repositories::TicketRepository.new.tickets_for_path(args[:feature_review_path])
+
+    has_links = tickets.any? { |ticket| ticket.key == args[:jira_key] }
+    return fail :missing_key, message: missing_key_message(args) unless has_links
+    continue(args)
+  end
+
+  def post_unlink_comment(args)
+    comment = self.class.build_comment("#{args[:root_url].chomp('/')}#{args[:feature_review_path]}")
+    JiraClient.post_comment(args[:jira_key], comment)
+    continue(success_message(args))
+  rescue JiraClient::InvalidKeyError
+    fail :invalid_key, message: invalid_key_message(args)
+  rescue StandardError => error
+    Honeybadger.notify(error)
+    fail :post_failed, message: generic_error_message(args)
+  end
+
+  private
+
+  def invalid_key_message(args)
+    "Failed to unlink #{args[:jira_key]}. Please check that the ticket ID is correct."
+  end
+
+  def missing_key_message(args)
+    "Failed to unlink #{args[:jira_key]}. Existing link couldn't be found."
+  end
+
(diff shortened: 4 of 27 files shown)


sldblog:
What are the plans for following this up? (So that people are aware)

MiroslavCsonka:
@sldblog We are going to fix it as a part of our "Let's fix it" experiment. https://jira.fundingcircle.com/browse/NS-1506

app/use_cases/unlink_ticket.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+  end
+
+  include SolidUseCase
+
+  steps :validate_id_format,
+        :assert_linked,
+        :post_unlink_comment
+
+  def validate_id_format(args)
```
MiroslavCsonka: DRY it up (same implementation as in `LinkTicket` use case)

spec/use_cases/unlink_ticket_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      context 'when posting is successful' do
+        let(:expected_message) {
+          "Feature Review was unlinked from #{jira_key}."\
+    ' Refresh this page in a moment and the ticket will disappear.'
+        }
+        let(:ticket) { Ticket.new(paths: [feature_review_path(apps_with_versions)], key: jira_key) }
+
+        before do
+          expect(ticket_repository).to receive(:tickets_for_path).and_return([ticket])
```
shabankarumba: Why don't you use `allow` as having an expectation does not look right here

spec/use_cases/unlink_ticket_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+        end
+
+        it 'sends a message to Jira' do
+          expect(JiraClient).to receive(:post_comment).with(jira_key, a_kind_of(String))
+
+          result = UnlinkTicket.run(args)
+          expect(result).to be_a_success
```
shabankarumba: I would put these two expectations in a separate `it` block

spec/use_cases/unlink_ticket_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  context 'when invalid ticket ID format' do
+    let(:jira_key) { 'INVALID' }
+    let(:expected_message) {
+      "Failed to unlink #{jira_key}. Please check that the ticket ID is correct."
+    }
+
+    it 'fails validation' do
+      result = UnlinkTicket.run(args)
```
shabankarumba: You can use `described_class` here as the class name may change

spec/models/repositories/ticket_repository_spec.rb
```diff
@@ -292,29 +292,59 @@
...
 
-      it 'projects the ticket referenced in the JIRA comments for each repository' do
-        [
-          build(:jira_event, key: 'JIRA-1', comment_body: "Review #{url1}", created_at: times[0]),
-          build(:jira_event, key: 'JIRA-1', comment_body: "Review [FR link|#{url2}]", created_at: times[1]),
-          build(:jira_event, key: 'JIRA-1', comment_body: "Review #{url1}", created_at: times[2]),
-        ].each do |event|
-          repository.apply(event)
+      context 'and applying link event' do
+        let(:url1) { feature_review_url(app1: 'one') }
```
shabankarumba: How about `url_one`?

spec/models/repositories/ticket_repository_spec.rb
```diff
@@ -292,29 +292,59 @@
...
+        let(:url2) { feature_review_url(app2: 'two') }
+        let(:path1) { feature_review_path(app1: 'one') }
```
shabankarumba: `path_one` ?

app/models/repositories/ticket_repository.rb
```diff
@@ -43,10 +142,17 @@ def apply(event)
 
       return if last_ticket.empty? && feature_reviews.empty?
 
-      new_ticket = build_ticket(last_ticket, event, feature_reviews)
+      new_ticket = if event.comment.include?(RelinkTicketJob::LINK_TICKET_EVENT_LABEL)
+                     RelinkTicketEvent.new(last_ticket, event).diff
+                   elsif event.comment.include?(UnlinkTicket::UNLINK_FEATURE_ACTION_LABEL)
```
bliof: 🤕 

So..
1. post comment to jira from `UnlinkTicket`
2. jira post event to here and based on the comment -> unlink / relink / link

ps. How is this working 💭

MiroslavCsonka: Yep that's exactly how it works. I just pushed a new version to simplify the event prrocessing. I didn't like the fact that `Repositories` were applying those events

Now it lives in `app/models/events/jira_event.rb`
snippet:
```ruby
     def handler
        if link_action?
          Handlers::LinkTicketHandler
        elsif unlink_action?
          Handlers::UnlinkTicketHandler
        else
          Handlers::TicketHandler
        end
      end
```

app/models/repositories/ticket_repository.rb
```diff
@@ -22,6 +22,7 @@ def tickets_for_path(feature_review_path, at: nil)
       store
         .select('DISTINCT ON (key) *')
         .where('paths @> ARRAY[?]', feature_review_path)
+        .where('id IN (select max(id) from tickets group by key)')
```
MiroslavCsonka: @bliof @0mega Do you know how this subquery could be removed? 

### The problem
Since we do not remove records in our projections (model Ticket), whenever we query our projections we also look through outdated records. 

In this case, imagine a ticket linked to a feature review `a` and `b`, when you want to unlink `b`, you just create a new ticket that has only `a` in it. But the query without subquery (`.where('id IN (select max(id) from tickets group by key)')`) will also look through outdated records which we don't want.

### My solutions so far

1. Leave the subquery in since this is the only place so far where it's an issue
1. Introduce `superseded_at` column to filterout outdated data
1. Start removing old records

My favourite is 1st because it's low impact, but I'm not sure what's the reason behind not doing 3rd from begining? Why do we need to store old tickets? It's just a projection derived from events, as long as we are not removing events we can always recreate ticket however we want.

bliof: Isn't this the same query?

```
shipment_tracker=> SELECT DISTINCT ON (key) id FROM "tickets" WHERE (paths @> ARRAY['/feature_reviews?apps%5Bsalesforce_metadata_uk%5D=18c9e409f7b3a20704e78c8488400022b88372c9']) AND (id IN (select max(id) from tickets group by key))  ORDER BY key, id DESC;
  id
-------
 56715
 56716
 56717
 56718
 56719
 56720
 56721
 56722
 56723
(9 rows)

shipment_tracker=> SELECT DISTINCT ON (key) id FROM "tickets" WHERE (paths @> ARRAY['/feature_reviews?apps%5Bsalesforce_metadata_uk%5D=18c9e409f7b3a20704e78c8488400022b88372c9'])  ORDER BY key, id DESC;
  id
-------
 56715
 56716
 56717
 56718
 56719
 56720
 56721
 56722
 56723
(9 rows)
```

where the old one is faster:

```
shipment_tracker=> explain SELECT DISTINCT ON (key) id FROM "tickets" WHERE (paths @> ARRAY['/feature_reviews?apps%5Bsalesforce_metadata_uk%5D=18c9e409f7b3a20704e78c8488400022b88372c9']) AND (id IN (select max(id) from tickets group by key))  ORDER BY key, id DESC;
                                                                  QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------
 Unique  (cost=7322.61..7322.68 rows=3 width=12)
   ->  Sort  (cost=7322.61..7322.64 rows=14 width=12)
         Sort Key: tickets.key, tickets.id
         ->  Nested Loop  (cost=6672.44..7322.34 rows=14 width=12)
               ->  HashAggregate  (cost=6672.15..6674.15 rows=200 width=4)
                     Group Key: max(tickets_1.id)
                     ->  HashAggregate  (cost=6560.98..6610.39 rows=4941 width=12)
                           Group Key: tickets_1.key
                           ->  Seq Scan on tickets tickets_1  (cost=0.00..6274.32 rows=57332 width=12)
               ->  Index Scan using tickets_pkey on tickets  (cost=0.29..3.23 rows=1 width=12)
                     Index Cond: (id = (max(tickets_1.id)))
                     Filter: (paths @> '{/feature_reviews?apps%5Bsalesforce_metadata_uk%5D=18c9e409f7b3a20704e78c8488400022b88372c9}'::text[])
(12 rows)

shipment_tracker=> explain SELECT DISTINCT ON (key) id FROM "tickets" WHERE (paths @> ARRAY['/feature_reviews?apps%5Bsalesforce_metadata_uk%5D=18c9e409f7b3a20704e78c8488400022b88372c9'])  ORDER BY key, id DESC;
                                                                    QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------
 Unique  (cost=1915.08..1915.23 rows=3 width=12)
   ->  Sort  (cost=1915.08..1915.16 rows=29 width=12)
         Sort Key: key, id
         ->  Bitmap Heap Scan on tickets  (cost=1804.22..1914.38 rows=29 width=12)
               Recheck Cond: (paths @> '{/feature_reviews?apps%5Bsalesforce_metadata_uk%5D=18c9e409f7b3a20704e78c8488400022b88372c9}'::text[])
               ->  Bitmap Index Scan on index_tickets_on_paths  (cost=0.00..1804.21 rows=29 width=0)
                     Index Cond: (paths @> '{/feature_reviews?apps%5Bsalesforce_metadata_uk%5D=18c9e409f7b3a20704e78c8488400022b88372c9}'::text[])
(7 rows)
```

bliof:
It will make more sense if we allow unlinking only if there isn't a deploy or maybe even a merge.

👍  / 👎  ?

app/models/repositories/ticket_repository.rb
bliof: ps. not sure how it exactly works but you can view feature reviews at a specific point in time:

hover on the Approve https://shipment-tracker.fundingcircle.co.uk/releases/shipment_tracker?region=gb

https://shipment-tracker.fundingcircle.co.uk/feature_reviews?apps%5Bshipment_tracker%5D=f563dfdb85879066f117fcece7d54dade7f8171b&time=2017-01-12+16%3A22%3A41+UTC

anujbiyani:
> It will make more sense if we allow unlinking only if there isn't a deploy or maybe even a merge.

Does make more sense, but I think that can be done as a separate chunk of work. @MiroslavCsonka?

bliof:
@anujbiyani 

It should simplify things. I am not sure how the current PR works when viewing a feature review at a specific time. (https://github.com/FundingCircle/shipment_tracker/pull/192#discussion_r99169785)

If we add the check in the UnlinkTicket use case all the hustle because of that ⬆️  will be gone. + We have to allow unlinking if you are looking only at the latest version of the feature review. 🤕  😕

0mega:
@bliof About unlinking a ticket for deployed feature review. I think it's ok if we don't do anything because in the moment that we received the DeployEvent, the state of the snapshots in the system had the approved ticket for the deployed feature review. What happens next to that ticket is not relevant as we present the feature review for a specific point in time.
But if I'm missing something please raise it :)

MiroslavCsonka:
@bliof : 
> how the current PR works when viewing a feature review at a specific time

That's an excellent point. I've completely missed the "time travel" feature. The query will need to be changed a bit.

bliof:
@MiroslavCsonka You'll disable it for time travelers, right? - hiding the button and adding a validation in `UnlinkTicket`

@0mega 👍 ps. https://github.com/FundingCircle/shipment_tracker/issues/203

MiroslavCsonka:
@bliof We've fixed the time travelling issue. And we are leaving the hiding for another ticket since we also need to hide "link ticket" form.

♻️  all the comments, so have another look

app/models/events/handlers/ticket_handler.rb
```diff
@@ -0,0 +1,39 @@
...
+      def feature_reviews_from_event
+        Factories::FeatureReviewFactory.new.create_from_text(event.comment)
+      end
+
+      private
+
+      def merge_approved_at(last_ticket, event)
+        if event.approval?
+          event.created_at
+        elsif last_ticket.present? && Ticket.new(status: event.status).approved?
```
shabankarumba: What if an event or last_ticket does not meet the conditions?

MiroslavCsonka: The value will be `nil` same as [previous implementation](https://github.com/FundingCircle/shipment_tracker/blob/master/app/models/repositories/ticket_repository.rb#L99).

app/use_cases/link_ticket.rb
```diff
@@ -4,25 +4,30 @@
 
 class LinkTicket
   include SolidUseCase
+  include TicketValidationHelper
+
+  COMMENT_LABEL = 'Feature ready for review'
```
shabankarumba: `freeze` ?

spec/use_cases/unlink_ticket_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+  before do
+    allow(Repositories::TicketRepository).to receive(:new).and_return(ticket_repository)
+    allow(ticket_repository).to receive(:tickets_for_path).and_return([])
+    allow(JiraClient).to receive(:post_comment)
+  end
+
+  describe '.class' do
+    subject(:unlink_comment) { described_class.run(args) }
+
+    context 'when invalid ticket ID format' do
```
shabankarumba: When the ticket ID is an invalid format

spec/use_cases/unlink_ticket_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      }
+
+      it 'fails validation' do
+        result = subject
+        expect(result).to fail_with(:invalid_key)
+        expect(result.value.message).to eq(expected_message)
+      end
+    end
+
+    context 'when valid ticket ID format' do
```
shabankarumba: When the ticket ID is a valid format

spec/use_cases/unlink_ticket_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      end
+
+      context 'when the ticket is linked' do
+        context 'when posting returns an error' do
+          let(:expected_message) { "Failed to unlink #{jira_key}. Something went wrong." }
+          let(:ticket) { Ticket.new(paths: [feature_review_path(apps_with_versions)], key: jira_key) }
+
+          before do
+            allow(JiraClient).to receive(:post_comment).and_raise(StandardError)
+            expect(ticket_repository).to receive(:tickets_for_path).and_return([ticket])
```
shabankarumba: Shouldn't this be an `allow` unless you want to assert on the `ticket_repository` receiving the message in that case you put in an `it` block ?

app/models/repositories/ticket_repository.rb
```diff
@@ -22,7 +22,7 @@ def tickets_for_path(feature_review_path, at: nil)
       store
         .select('DISTINCT ON (key) *')
```
bliof: This is most probably not needed.

app/models/repositories/ticket_repository.rb
```diff
@@ -31,6 +31,7 @@ def tickets_for_versions(versions)
...
```
bliof: and this one

app/models/repositories/ticket_repository.rb
```diff
@@ -22,7 +22,7 @@ def tickets_for_path(feature_review_path, at: nil)
...
         .where('paths @> ARRAY[?]', feature_review_path)
-        .where(query)
+        .where(id: store.select('MAX(id) as id').where(query).group(:key))
```
bliof: Miroslav Csonka:

> The problem is when you have a ticket and it's linked to feature reviews 1,2,3, and when you unlink 2. In this case when you go to page for 1,2,3 it would pull up the outdated jira ticket and show you page that's no longer true.

because of the `.where('paths @> ARRAY[?]', feature_review_path)`

spec/models/events/handlers/link_ticket_handler_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  describe '#apply' do
+    it 'adds link references' do
+      events = [
+        build(:jira_event, key: 'JIRA-1', comment_body: LinkTicket.build_comment(url1), created_at: time1),
+        build(:jira_event, key: 'JIRA-1', comment_body: LinkTicket.build_comment(url2), created_at: time2),
+        build(:jira_event, key: 'JIRA-1', comment_body: LinkTicket.build_comment(url1), created_at: time - 1.hour),
+      ]
+
+      final_ticket = events.reduce({}) { |ticket, event| described_class.new(ticket, event).apply }
+      expect(final_ticket).to match(hash_including(
```
bliof: How about:

```
expect(final_ticket).to match(
  hash_including(
    'key' => 'JIRA-1',
    'paths' => [path1, path2],
    'versions' => %w(one two),
    'version_timestamps' => { 'one' => time1, 'two' => time2 },
  ),
)
```

spec/use_cases/unlink_ticket_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+          it 'fails with message' do
+            expect(subject).to fail_with(:post_failed)
+            expect(subject.value.message).to eq(expected_message)
+          end
+        end
+
+        context 'when posting is successful' do
+          let(:expected_message) {
+            "Feature Review was unlinked from #{jira_key}."\
+    ' Refresh this page in a moment and the ticket will disappear.'
```
bliof: :fist_right: ➡️ ➡️

spec/models/events/handlers/unlink_ticket_handler_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe Events::Handlers::UnlinkTicketHandler do
+  let(:time) { Time.current.change(usec: 0) }
+
+  let(:url1) { feature_review_url(app1: 'one') }
```
bliof: This is not used?

spec/models/events/handlers/unlink_ticket_handler_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  let(:path1) { feature_review_path(app1: 'one') }
+  let(:time1) { time - 3.hours }
```
bliof: Let's move (remove) the `time` related `let`s inside the it block

app/use_cases/link_ticket.rb
MiroslavCsonka: Freezing is no longer necessary thanks to `# frozen_string_literal: true` (see top of the file)

PR #194, opened 2016-12-20
GQ-28-Add repository owners

bliof:
Add an option to specify "Repo Owner" to a repository.

To add an repo owner you specify either:
 * `test@foo.com` - will search for a project owner with that email or create one
 * `John Snow <test@foo.com>` - will search with the email and create / update the project owner by setting also his name

You can specify multiple repo owners like this:

```
repo-owner1@test.com, repo-owner2@test.com
repo-owner3@test.com
```

- [x] Add features specs

### DIFF
--- a/app/models/git_repository_location.rb
+++ b/app/models/git_repository_location.rb
@@ -4,7 +4,7 @@ require 'git_clone_url'
 class GitRepositoryLocation < ActiveRecord::Base
   before_validation on: :create do
     self.uri = uri&.strip
-    self.name = extract_name(uri)
+    self.name ||= extract_name(uri)
   end
 
   validates :uri, presence: true
@@ -49,7 +49,15 @@ class GitRepositoryLocation < ActiveRecord::Base
     end
   end
 
+  def owners
+    repo_ownership_repository.owners_of(self)
+  end
+
   private
+
+  def repo_ownership_repository
+    Repositories::RepoOwnershipRepository.new
+  end
 
   def self.url_from_uri(uri)
     parsed_uri = GitCloneUrl.parse(uri)
--- a/app/models/repositories/updater.rb
+++ b/app/models/repositories/updater.rb
@@ -14,10 +14,14 @@ module Repositories
       @repositories = repositories
     end
 
-    def run(repo_event_id_hash = {})
+    def recreate(repo_event_id_hash = {})
+      reset
+      run(repo_event_id_hash, true)
+    end
+
+    def run(repo_event_id_hash = {}, force = false)
       repositories.each do |repository|
-        Rails.logger.info "[#{Time.current}] Running recreate_snapshots for repo #{repository.table_name}"
-        run_for(repository, repo_event_id_hash[repository.table_name])
+        run_for(repository, repo_event_id_hash[repository.table_name], force)
       end
     end
 
@@ -33,15 +37,33 @@ module Repositories
 
     attr_reader :repositories
 
-    def run_for(repository, ceiling_id)
-      repository.store.transaction do
-        last_id = 0
-        new_events_for(repository, ceiling_id).each do |event|
-          last_id = event.id
-          repository.apply(event)
+    def run_for(repository, ceiling_id, force = false)
+      if force || run_in_background?(repository)
+        Rails.logger.info "[#{Time.current}] Apply events for #{repository.class} - #{repository.table_name}"
+
+        repository.store.transaction do
+          last_id = 0
+
+          new_events_for(repository, ceiling_id).each do |event|
+            last_id = event.id
+            repository.apply(event)
+          end
+
+          update_count_for(repository, last_id) unless last_id == 0
         end
-        update_count_for(repository, last_id) unless last_id == 0
+      else
+        Rails.logger.info(
+          "[#{Time.current}] Skip applying events for #{repository.class} - #{repository.table_name}",
+        )
       end
+    end
+
+    def run_in_background?(repository)
+      repositories_that_do_not_run_in_the_background = [
+        Repositories::RepoOwnershipRepository,
+      ]
+
+      !repositories_that_do_not_run_in_the_background.include?(repository.class)
     end
 
     def new_events_for(repository, ceiling_id)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -10,4 +10,12 @@ class User
   def logged_in?
     email.present?
   end
+
+  def owner_of?(repository)
+    logged_in? && as_repo_owner.owner_of?(repository)
+  end
+
+  def as_repo_owner
+    @repo_owner ||= RepoOwner.find_by(email: email) || RepoOwner.new(email: email, name: first_name)
+  end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,23 +1,14 @@
-# Files in the config/locales directory are used for internationalization
-# and are automatically loaded by Rails. If you want to use locales other
-# than English, add the necessary files in this directory.
-#
-# To use the locales, use `I18n.t`:
-#
-#     I18n.t 'hello'
-#
-# In views, this is aliased to just `t`:
-#
-#     <%= t('hello') %>
-#
-# To use a different locale, set it with `I18n.locale`:
-#
-#     I18n.locale = :es
-#
-# This would use the information in config/locales/es.yml.
-#
-# To learn more, please read the Rails Internationalization guide
-# available at http://guides.rubyonrails.org/i18n.html.
-
 en:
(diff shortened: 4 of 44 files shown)


popovm:
Hey @0mega, could you take a look at this pull request 😃

app/use_cases/change_repository_project_owner.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+class ChangeRepositoryProjectOwner
```
0mega: Can we have the UseCases inheriting from SolidUseCase as in the existing ones? e.g.: https://github.com/FundingCircle/shipment_tracker/blob/master/app/use_cases/add_repository_location.rb#L5

app/use_cases/remove_repository_project_owner.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+class RemoveRepositoryProjectOwner
```
0mega: Inherit from SolidUseCase

config/locales/en.yml
```diff
@@ -1,23 +1,11 @@
...
-#
-# To learn more, please read the Rails Internationalization guide
-# available at http://guides.rubyonrails.org/i18n.html.
-
 en:
-  hello: "Hello world"
+  activemodel:
+    attributes:
+      'forms/edit_git_repository_location_form':
+        project_owner: 'Repo/PO/Dev Owner'
```
0mega: Can we change 'Repo/PO/Dev Owner'? Maybe it's better to call it just 'Repo Owners'.

app/models/git_repository_location.rb
```diff
@@ -10,6 +10,8 @@ class GitRepositoryLocation < ActiveRecord::Base
   validates :uri, presence: true
   validates :name, uniqueness: true
 
+  belongs_to :project_owner
```
0mega: If I remember correctly @fernandezj requested support for multiple project owners. So we probably need a has_many relation here.

Another option would be to not create a ProjectOwner model and save the owners in a JSON field of GitRepositoryLocation. The JSON stored in `project_owners` column could be like:
```
{{name: Me, email: me@test.com}, {name:'', email: 'you@test.com}}
```
AFAIK project owners will only be used for sending emails whenever there is a deploy alert.

Another question, I'm not against creating a dedicated model as long as we update the relation, but I was wondering if you have considered using the existing `User` model instead of creating a new one, can you see any inconvenient in that?

app/use_cases/change_repository_project_owner.rb
bliof: I think we should drop this gem. I'll open an issue to describe why but the tl&dr; is we do not benefit from it and there is hardly anything that it can give us if we decide to use it fully.

config/locales/en.yml
bliof: This was suggested in the jira. It looks buggy but it is short and catches a larger scope than "Repo Owners" ¯\_(ツ)_/¯

app/models/git_repository_location.rb
bliof: Most probably it should be habtm with soft delete so we do not loose any historical data.

👎  on using json

They will receive emails for alerts and have the ability approve/explain unapproved deployments by writing a comment. This should be all for the moment.

As far as I get it there a some light models e.g. `User` model which are initialized with data from somewhere else (e.g. database, oauth, random event). So.. in this PR the store is the `ProjectOwner` and we use the `User` model by adding `#project_owner_of?` method.

config/locales/en.yml
steveweet: I agree with @0mega the phrase 'Repo/PO/Dev Owner', whilst agreed upon in the Jira conversatio, adds confusion rather than clarifying. I would go with the more specific 'Repo Owners'

app/helpers/repo_owner_helper.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+module RepoOwnerHelper
+  def repo_owner_name_and_email(repo)
```
bliof: - [ ] we should pass the repo_owner here

app/models/user.rb
```diff
@@ -10,4 +10,8 @@ class User
   def logged_in?
     email.present?
   end
+
+  def repo_owner_of?(repository)
```
0mega: Not a blocker: suggest simplifying to
```ruby
def owner_of?(repository)
```

app/helpers/repo_owner_helper.rb
0mega: 👍 , and it should probably be a list of repo_owners

app/models/git_repository_location.rb
0mega: Agree on using habtm instead of JSON if we are using the ProjectOwner/RepoOwner for other use cases.

It looks like we should also make an association of ProjectOwner and User. Maybe Project Owner belongs to User?

> Which begs the question of whether we should access the name/email from the user model itself..

If we use the `email` and `name` from User model, than we have a problem of not knowing if that specific user exists given that the user must have had logged in to ShipmentTracker before so that it's added to the DB. But it does not sound like a big problem (we just ask the specific person to login).

bliof:
@0mega 👓

features/repository_locations.feature
```diff
@@ -23,3 +23,24 @@ Scenario: Add repository with auto-generated tokens
     | Source               | Name     | Endpoint        |
     | CircleCI (webhook)   | app_name | circleci        |
     | Deployment           | app_name | deploy          |
+
+@disable_repo_verification
+Scenario: Add an owners of a repository
```
0mega: Small thing: "Add an owner of a repository" or "Add owners of a repository"

bliof:
@0mega Review ? :) 👓 

Any idea if there is a better way to handle the uri for the application in the features spec?

It is not really clear what "Repository"/"Repo" is.. I was using "Repo" for git repository location, because "Repository" is used to note the pattern. But in ScenarioContext there is some other "Repo" 🙊

bliof:
I think we’ve settled on how the shipment tracker is structured:

It is split in two part:
  1. the repos (name, git uri)
  2. metadata for repos

The metadata is the part with the events/snapshots/repositories (the thing that applies an event)
There is a table called events and it holds serialised data for all of them.

The idea is that from that table you could build all the metadata. This does not include the repos.
You could apply all the events and add a repo for bilcas and you'll see only only the data for bilcas.
In config/initializers/repositories.rb you specify in what order the events should be applied on recovery

events use the name of the app to link to repos

```
                     apply                               will create
RepoOwnershipEvent ---------> RepoOwnershipRepository ------------------> Snapshot::RepoOwnership
                                   |                                |---> RepoOwner
                    owners_of_repo |
[RepoOwner*]      <----------------|
```

Snapshot::RepoOwnership
 - app name
 - emails of the owners

So.. to get the repo owners metadata you'll load the last snapshot for a repo with certain name and
then load the data from RepoOwner which we'll mainly use to keep the names of the owners insync

For the DeployAlerts tracking we'll have to add a uuid for all current events because the relationship between the
DeployEvent and DeployAlertEvent is 1-1, both events will update the Snapshot::Deploy

Usually for the snapshots there is a value object e.g. models/deploy.rb

---

I'll update this PR shortly

bliof:
@0mega @popovm 👓 
@popovm You can rebase and deploy to heroku.

bliof:
ps there are some small fixes eg:
- `EventFactory#create` will do `create!`
- There is a new `form_errors` helper
- Also the form for adding repositories/projects

app/models/repositories/updater.rb
```diff
@@ -33,17 +37,35 @@ def reset
...
+        Rails.logger.info(
+          "[#{Time.current}] Skip applying events for #{repository.class} - #{repository.table_name}",
+        )
       end
     end
 
+    def run_in_background?(repository)
+      repositories_that_do_not_run_in_the_background = [
+        Repositories::RepoOwnershipRepository,
+      ]
```
0mega: Thinking again, I guess the best thing would be to get the EventType for the repo that is being analysed and check if the `internal` property is set to true, and if it is, apply this logic.
Could be something like:
```ruby
def is_external?(repository)
  event_type = get_type(repository)
  !event_type.internal
end
```
And than you would replace the `if force || run_in_background?(repository)` on line 41 with `if force || is_external?(repository)`. This way you don't have to make another "special" list here

bliof: There is the `manual_test` event type which is also internal. It should be the one for QA comments.

Let's leave it as is and do the refactoring later when we have also the form with the exception comments.

PR #197, opened 2017-01-03
Spec refactoring

bliof:
* spec/jobs/deploy_alert_job_spec.rb
* spec/models/repositories/deploy_repository_spec.rb


bliof:
@0mega 👀

PR #199, opened 2017-01-12
Fix update git loop rake task

vnegrisolo:
- [x] update test to prove the bug on the query
- [x] the query was returning the first match, and the result is unpredictable
- [x] fix the bug, change the ILKE criteria do look for "exact" repo name
- [x] use pluck to build a hash from database for perf

In summary, `FundingCircle/bilcas` was conflicting with:
- FundingCircle/bilcas-client
- FundingCircle/bilcas-cookbook
- FundingCircle/bilcas_sub
- FundingCircle/bilcas_adapter

**JIRA issue**: https://jira.fundingcircle.com/browse/CASS-992

### DIFF
--- a/spec/models/git_repository_location_spec.rb
+++ b/spec/models/git_repository_location_spec.rb
@@ -35,15 +35,12 @@ RSpec.describe GitRepositoryLocation, :disable_repo_verification do
   end
 
   describe '.app_remote_head_hash' do
-    let(:app_locations) {
-      [
-        GitRepositoryLocation.new(name: 'app1', remote_head: 'abc123'),
-        GitRepositoryLocation.new(name: 'app2', remote_head: 'abc456'),
-      ]
-    }
+    before do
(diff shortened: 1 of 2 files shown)


spec/models/git_repository_location_spec.rb
```diff
@@ -108,6 +105,7 @@
   describe '.find_by_full_repo_name' do
     context 'when a GitRepositoryLocation exists with the same name' do
       it 'returns a GitRepositoryLocation' do
+        GitRepositoryLocation.create(uri: 'git@github.com:owner/repo-frontend.git')
```
0mega: Wondering if you add:
```ruby
GitRepositoryLocation.create(uri: 'git@github.com:owner/old-repo-frontend.git')
```
will the test still pass? 🤔

vnegrisolo: based on the tests I think that it will, because we should receive `owner/repo`

vnegrisolo: I added to the test just in case and as a documentation of this question

PR #200, opened 2017-01-20
Add guard

bliof:
* https://github.com/guard/guard-cucumber/tree/v1.5.4
* https://github.com/guard/guard-rspec/tree/v4.7.3

spring rspec - does not work very well because of some errors with the
autoloading. This will be resolved in another PR.

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+guard(
+  :rspec,
+  cmd: 'bundle exec rspec',
+  run_all: { cmd: 'bundle exec rspec', message: 'Running all specs' },
+) do
+  require 'guard/rspec/dsl'
+
+  dsl = Guard::RSpec::Dsl.new(self)
+
+  # Feel free to open issues for suggestions and improvements
+
+  # RSpec files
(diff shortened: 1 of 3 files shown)


bliof:
@popovm @0mega 👀  Let's add guard 🛡

Guardfile
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+guard(
+  :rspec,
+  cmd: 'spring rspec',
```
bliof: ... this does not work all the time 😢

0mega: what do you mean, the tests fail?

bliof: There are some issues with the autoloading

```
<home>/.rvm/gems/ruby-2.3.0@shipment_tracker/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:495:in `load_missing_constant': Unable to autoload constant Events::CircleCiEvent, expected <home>/projects/shipment_tracker/app/models/events/circle_ci_event.rb to define it (LoadError)
```

popovm:
LGTM 🐎

PR #201, opened 2017-01-27
Track unapproved deployments

bliof:
The PR aims to record the deploy alerts in the deploy snapshots

```
DeployAlertEvent
  - deploy snapshot uuid
  - message

Snapshots::Deploy
  - deploy_alert - string
```

To do this a uuid is added to all events. The uuid from the deploy event will be saved in the deploy snapshot. It will be used to refer to the snapshot in a later stage from the deploy alert event.

Old deploy snapshot for the moment will not have a uuid. Maybe they could be migrated based on the timestamps :man_shrugging: 

Events are now applied in order. New repositories will not apply all the old
events from the start by default. You'll have to do a recreate manually.

Add Repositories::Base - all repositories inherit from this one

Use global event pointer and do not use the per repository event counts
The code for tracking events per repository is still there. Most
probably it could be removed. Only the global event pointer is used in
the updater.

### DIFF
--- a/app/models/repositories/updater.rb
+++ b/app/models/repositories/updater.rb
@@ -14,14 +14,24 @@ module Repositories
       @repositories = repositories
     end
 
-    def recreate(repo_event_id_hash = {})
+    attr_accessor :repositories
+
+    def recreate(upto_event: nil)
       reset
-      run(repo_event_id_hash, true)
+      run(apply_to_all: true, upto_event: upto_event)
     end
 
-    def run(repo_event_id_hash = {}, force = false)
-      repositories.each do |repository|
-        run_for(repository, repo_event_id_hash[repository.table_name], force)
+    def run(apply_to_all: false, upto_event: nil)
+      repositories_to_use = filter_repositories(apply_to_all: apply_to_all)
+
+      new_events(upto_event: upto_event).each do |event|
+        ActiveRecord::Base.transaction do
+          repositories_to_use.each do |repository|
+            apply repository: repository, event: event
+          end
+
+          Snapshots::EventCount.global_event_pointer = event.id
+        end
       end
     end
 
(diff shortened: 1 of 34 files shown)


config/initializers/repositories.rb
```diff
@@ -12,4 +12,5 @@
   # UatestRepository must always be last as it depends on DeployRepository.
```
popovm: 👮

spec/jobs/deploy_alert_job_spec.rb
```diff
@@ -38,13 +50,57 @@
...
+      described_class.perform_now(
+        current_deploy: deploy_attrs(
           'app_name' => 'frontend',
           'deployed_by' => 'John',
           'region' => 'uk',
         ),
       )
     end
+
+    it 'creates a deploy alert event if there is a alert message' do
```
popovm: an alert message 👴

bliof:
@popovm @0mega 👀  Do you agree with such a change in the `Repositories::Updater`?

I am not sure why I did it, I "remembered" just now that only the only the `Repositories::Base` was needed to apply issue - DeployRepository and DeployAlertRepository use the same store => same table_name.

Todo:
  - [ ] add few more specs
  - [ ] decide whether to keep the event counts per repository
       - [ ] custom identifiers for repositories the table name is not that good
  - [ ] recreate the database from the events (maybe try on a test database first)

app/models/snapshots/event_count.rb
```diff
@@ -3,8 +3,30 @@
 
 module Snapshots
   class EventCount < ActiveRecord::Base
-    def self.repo_event_id_hash
-      pluck(:snapshot_name, :event_id).to_h
+    class << self
+      def global_event_pointer
+        find_by_snapshot_name('global_event_pointer')&.event_id || 0
```
popovm: Nitpick: That could be written in the same way as `pointer_for`

app/models/repositories/updater.rb
```diff
@@ -14,14 +14,24 @@ def initialize(repositories)
       @repositories = repositories
     end
 
-    def recreate(repo_event_id_hash = {})
+    attr_accessor :repositories
+
+    def recreate(upto_event: nil)
       reset
-      run(repo_event_id_hash, true)
+      run(apply_to_all: true, upto_event: upto_event)
```
popovm: `up_to_event` or `until_event`?

popovm:
What if we put the global pointer to the newest event for now? When we are confident that the database could be recreated safely, we would recreate it and get the historic deployment alerts?

app/models/repositories/updater.rb
```diff
@@ -14,14 +14,24 @@ def initialize(repositories)
...
-    def run(repo_event_id_hash = {}, force = false)
-      repositories.each do |repository|
-        run_for(repository, repo_event_id_hash[repository.table_name], force)
+    def run(apply_to_all: false, upto_event: nil)
+      repositories_to_use = filter_repositories(apply_to_all: apply_to_all)
+
+      new_events(upto_event: upto_event).each do |event|
+        ActiveRecord::Base.transaction do
+          repositories_to_use.each do |repository|
+            apply repository: repository, event: event
```
0mega: Not a blocker but I think we follow the convention of explicit method invocation `apply(repository: repository, event: event)`

app/models/repositories/updater.rb
```diff
@@ -35,51 +45,29 @@ def reset
...
-    end
-
-    def run_in_background?(repository)
-      repositories_that_do_not_run_in_the_background = [
-        Repositories::RepoOwnershipRepository,
-      ]
-
-      !repositories_that_do_not_run_in_the_background.include?(repository.class)
+    def apply(repository:, event:)
+      Rails.logger.info "[#{Time.current}] Apply events for #{repository.class} - #{repository.table_name}"
```
0mega: s/events/event

db/migrate/20170127123616_add_deploy_alert_to_deploys.rb
```diff
@@ -0,0 +1,6 @@
+class AddDeployAlertToDeploys < ActiveRecord::Migration
+  def change
+    add_column :deploys, :deploy_alert, :string
+    add_index :deploys, :deploy_alert
```
bliof: We are going to add some type of audit report => we'll need all deploys that have alerts.

lib/tasks/jobs.rake
```diff
@@ -55,11 +52,13 @@ namespace :jobs do
...
 
         Repositories::Updater.from_rails_config.run
 
-        num_events = Events::BaseEvent.where('id > ?', lowest_event_id).count
+        last_event_id = Snapshots::EventCount.global_event_pointer
+
+        num_events = Events::BaseEvent.where('id > ?', from_event_id).where('id <= ?', last_event_id).count
 
         end_time = Time.current
         puts "[#{end_time}] Cached #{num_events} events in #{end_time - start_time} seconds"
```
bliof: A lot of things are not part of the change :D

PR #204, opened 2017-02-03
Make sure that we have only events that can be applied by repositories

bliof:
When applying ReleaseExceptionEvent we validate that the repo owner has
access to the repo. When we recreate the database this will fail.

The validation should happen before we create the event so we can
recreate the database easily without having to deal with invalid events.


bliof:
@popovm 👀

PR #205, opened 2017-02-06
[PO-2748] Fix commit status in github for reviews with release exceptions

popovm:
https://jira.fundingcircle.com/browse/PO-2748

Currently status in github will not be updated if a release exception is created 😢

### DIFF
--- a/app/models/queries/decorated_feature_reviews_query.rb
+++ b/app/models/queries/decorated_feature_reviews_query.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+module Queries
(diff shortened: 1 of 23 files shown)


popovm:
Ping @bliof for comments :)

app/models/queries/decorated_feature_reviews_query.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+# require 'feature_review_with_statuses'
```
bliof: ✂️

PR #212, opened 2017-04-26
[PO-2994] Handle release exceptions not related to deploys

popovm:
If a release exception is corresponding to a pending deploy, then `Not Deployed` will be displayed on the report.


popovm:
@bliof 👀

PR #216, opened 2017-05-23
Consider Unknown versions in unapproved deploys report

popovm:
When a deployment notification is sent without specified commit SHA, the deployment is not associated with any software version. That leads to undefined method errors on the unapproved deployments page. 🙅 

Honeybadger link: https://app.honeybadger.io/projects/44071/faults/33406513

Release exception are currently not associated with deploys, because deploys are related to merge commit, but release exception to the previous commits. That's why there is no benefits at the moment to search for deploys related to release exceptions.


popovm:
@bliof 👀

PR #217, opened 2017-05-23
Force html format for unapproved deployments

popovm:
It's the same issue as the one fixed in #127 
The page for applications ending in `.js` is not rendered correctly at the moment.


popovm:
@bliof 👀

PR #218, opened 2017-06-07
[PO-2870] Improve alerts for unauthorised deploys

mbdimitrova:
When a deploy happens all other merges that happened since the last deployment are considered. If any of these have not been authorised, a list of their SHAs will be added in the deploy alert message.

Relates:
- [x] JIRA :gem:: https://jira.fundingcircle.com/browse/PO-2870

### DIFF
--- a/app/models/deploy_alert.rb
+++ b/app/models/deploy_alert.rb
@@ -21,12 +21,13 @@ class DeployAlert
     elsif deploy_auditor.rollback?
       alert_rollback(current_deploy)
     elsif !deploy_auditor.recent_releases_authorised?
-      alert_not_authorised(current_deploy)
+      alert_not_authorised(current_deploy, deploy_auditor)
     end
   end
 
-  def self.alert_not_authorised(deploy)
-    "#{alert_header(deploy)}Release not authorised; Feature Review not approved."
+  def self.alert_not_authorised(deploy, deploy_auditor)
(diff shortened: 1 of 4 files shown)


app/models/deploy_alert.rb
```diff
@@ -21,12 +21,13 @@ def self.audit_message(current_deploy, previous_deploy = nil)
...
       alert_rollback(current_deploy)
     elsif !deploy_auditor.recent_releases_authorised?
-      alert_not_authorised(current_deploy)
+      alert_not_authorised(current_deploy, deploy_auditor)
     end
   end
 
-  def self.alert_not_authorised(deploy)
-    "#{alert_header(deploy)}Release not authorised; Feature Review not approved."
+  def self.alert_not_authorised(deploy, deploy_auditor)
```
bliof: 🙋‍♂️ Let's leave only `#audit_message` and `#auditable?` visible.

app/views/unapproved_deployments/show.html.haml
```diff
@@ -44,7 +44,7 @@
           %div= feature_review_link(feature_review)
       %td= time_with_timezone(deploy.deployed_at)
       %td= deploy.deployed_by
-      %td= deploy.deploy_alert
+      %td= simple_format(deploy.deploy_alert)
```
bliof: 👏

PR #219, opened 2017-07-26
Remove the UAT environment forms + views

bliof:
## Change

At the moment a feature review is defined by the commit sha and a uat env on which it is tested. The current change drops the uat env and now a feature review will be defined only by the commit sha (or multiple commit sha-s).

- Deploys to uats are still tracked.
- The User Acceptance will be introduced again in a follow-up PR - the check will represent whether the change has passed automated acceptance tests
- QA Acceptance panel will also be modified in a follow-up PR - we'll keep all QA comments even after rebases

screen shot 2017-07-27 at 14 45 02

### UAT Environment

- [x] Remove UAT Environment from Summary.
- [x] Remove UAT Environment section.
- [x] Remove "No UAT specified" from Apps Under Review section.

### User Acceptance

- [x] Remove User Acceptance from Summary. 
- [x] Remove User Acceptance Panel and checks. 
- [x] Remove User Acceptance Testing UAT environment field the form for preparing/modifying feature reviews.
- [x] Remove `uat_url` from feature reviews.

### From where to start the review

Check changes to `uat_url`, `deploy_changes`, `deploys_for`.

The `deploys` for `feature_review_with_statuses` were actually deploys to uat.


app/models/queries/feature_review_query.rb
```diff
@@ -15,7 +14,6 @@ def initialize(feature_review, at:)
       @deploy_repository = Repositories::DeployRepository.new
```
popovm: This repository was used only in `#deploys`, so now is no longer needed

PR #220, opened 2017-08-01
Add integration test field on Feature Review page

popovm:
This pull request adds a field to the Feature Review page, which displays the status of the integration test build.
screen shot 2017-09-08 at 18 11 57

### DIFF
--- a/app/models/repositories/build_repository.rb
+++ b/app/models/repositories/build_repository.rb
@@ -37,7 +37,7 @@ module Repositories
     private
 
     def test_results_for(apps:, build_type:, at: nil)
-      default_builds = apps.keys.map { |app_name| [app_name, Build.new] }.to_h
+      default_builds = apps.keys.map { |app_name| [app_name, Build.new(build_type: build_type)] }.to_h
       builds = builds(apps.values, build_type, at).map { |build| [apps.invert[build.version], build] }.to_h
       default_builds.merge(builds)
     end
--- a/app/models/repositories/build_repository.rb
+++ b/app/models/repositories/build_repository.rb
@@ -6,33 +6,46 @@ require 'snapshots/build'
 
 module Repositories
   class BuildRepository < Base
+    APPLICABLE_EVENTS = [Events::CircleCiEvent, Events::JenkinsEvent].freeze
+
     def initialize(store = Snapshots::Build)
       @store = store
     end
 
     def apply(event)
-      return unless event.is_a?(Events::CircleCiEvent) || event.is_a?(Events::JenkinsEvent)
+      return unless APPLICABLE_EVENTS.any? { |event_type| event.is_a? event_type }
 
       store.create!(
         success: event.success,
+        app_name: event.try(:app_name),
+        build_type: event.build_type,
         source: event.source,
         version: event.version,
+        url: event.build_url,
         event_created_at: event.created_at,
       )
     end
 
-    def builds_for(apps:, at: nil)
-      default_builds = apps.keys.map { |app_name| [app_name, Build.new] }.to_h
-      builds = builds(apps.values, at).map { |build| [apps.invert[build.version], build] }.to_h
-      default_builds.merge(builds)
+    def integration_test_results_for(apps:, at: nil)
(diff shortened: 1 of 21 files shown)


popovm:
@FundingCircle/prod-ops @bliof @0mega  👀

app/models/events/integration_test_event.rb
```diff
@@ -0,0 +1,22 @@
...
+  class IntegrationTestEvent < Events::BaseEvent
+    def app_name
+      details.fetch('app_name')
+    end
+
+    def version
+      details.fetch('version', 'unknown')
+    end
+
+    def url
```
bliof: maybe `details_url` will be better

0mega:
@sgerrand there might be some confusion with what the name represents. My understanding was that the `IntegrationTest` will represent the tests that are run on Jenkins for our apps. So it's part of the core model of shipment tracker to represent other application's result of running the integration tests, as opposed to tests that we run against shipment tracker code.

sgerrand:
Whoops, apologies. 😊

app/models/integration_test.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+class IntegrationTest
```
sgerrand: Suggest `IntegrationTestResult` as a more explicit name.

popovm:
We discussed with @bliof and we agreed that actually there is no reason of a new event type.
We can re-use the `BuildEvent`, as this is actually a build and just have a new type of `source`, which states that this is a integration tests build.

popovm:
@bliof @0mega That's ready for review again 👀

app/models/repositories/build_repository.rb
```diff
@@ -6,22 +6,38 @@
...
 
-    def builds_for(apps:, at: nil)
+    def integration_tests_result_for(versions:, at: nil)
+      created_at_query = at ? store.arel_table['event_created_at'].lteq(at) : nil
+
+      store
+        .where(version: versions, build_type: 'integration')
+        .where(created_at_query)
+        .order('id DESC')
+        .first
```
bliof: Let's allow multiple builds like the one for unit tests.

popovm:
@mbdimitrova @bliof Could you take a final look at this pull request 👀

spec/decorators/feature_review_with_statuses_spec.rb
```diff
@@ -118,51 +123,84 @@
...
     end
+  end
 
-    context 'when there are no builds' do
-      let(:builds) { {} }
+  describe '#integration_test_result_status' do
+    context 'when all integration test results pass' do
+      let(:integration_test_results) do
+        {
+          'frontend' => Build.new(success: true),
```
bliof: It would be better to add a factory for `unit_test_build` and `integration_test_build`.

popovm: @bliof Unfortunately, `Build` is not a `ActiveRecord`, but a `Virtus` object. That's why I am not sure we can use factory_girl for building `builds`

popovm:
@bliof I added a `BuildHelper` for constructing builds.

spec/decorators/feature_review_with_statuses_spec.rb
sgerrand: Factory Girl can build any kind of Ruby object, including POROs. 😉

spec/support/build_helper.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Support
+  module BuildHelper
+    def build_unit_test_build(options = {})
```
sgerrand: 🤔  Please use a factory for this instead. It's simpler to understand how and where test objects are constructed if it all happens in the one place.

e.g.
```ruby
FactoryGirl.define do
  factory :build do
    factory :integration_test_build do
      build_type: 'integration'
    end

    factory :unit_test_build do
      build_type: 'unit'
    end
  end
end
```

popovm:
@sgerrand I added factories for builds and removed the previous helper. 🙂

PR #221, opened 2017-08-31
[PO-3612] QA text persistence

mbdimitrova:
When a Feature Review is generated, QA Submission messages for all related commits are shown. A related commit is every commit between the last merge commit and the given commit. This applies for all applications for which the Feature Review is generated:

screen shot 2017-08-31 at 15 29 03

Relates:
- [x] JIRA :gem:: https://jira.fundingcircle.com/browse/PO-3612

/cc @FundingCircle/prod-ops 🙇‍♀️

### DIFF
--- a/app/models/queries/feature_review_query.rb
+++ b/app/models/queries/feature_review_query.rb
@@ -44,7 +44,7 @@ module Queries
 
     def qa_submissions
       manual_test_repository.qa_submissions_for(
-        versions: feature_review.related_app_versions,
+        versions: feature_review.related_app_versions.values.flatten,
         at: time,
       )
     end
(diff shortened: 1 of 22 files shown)


app/models/queries/feature_review_query.rb
```diff
@@ -42,9 +42,9 @@ def builds
       )
     end
 
-    def qa_submission
-      manual_test_repository.qa_submission_for(
-        versions: feature_review.versions,
+    def qa_submissions
+      manual_test_repository.qa_submissions_for(
+        versions: feature_review.related_app_versions,
```
bliof: This should be
```
feature_review.related_app_versions.values.flatten
```

not it works because you do `{a: 123, b: 4325}.sort.join(', ') # => a, 123, b, 4325`

denislavski:
Codeclimate is complaints are not real duplications so we will bypass the check.

PR #222, opened 2017-09-08
[PO-3657] Make tests a required check for some repositories

mbdimitrova:
🚢 This pull request adds a new setting to repositories - **Required checks** 🛡  In order to have an authorised Feature Review, all selected (if any) required checks for the reviewed repositories must have passed :shipit: 

screen shot 2017-09-08 at 14 53 50

Relates:
- [x] JIRA :gem:: https://jira.fundingcircle.com/browse/PO-3657

/cc @FundingCircle/prod-ops, @0mega 🙇‍♀️

### DIFF
--- a/app/controllers/git_repository_locations_controller.rb
+++ b/app/controllers/git_repository_locations_controller.rb
@@ -31,7 +31,7 @@ class GitRepositoryLocationsController < ApplicationController
   def update
     load_edit_env
 
-    edit_params = params.require(:forms_edit_git_repository_location_form).permit(:repo_owners)
+    edit_params = params.require(:forms_edit_git_repository_location_form).permit(:repo_owners, required_checks: [])
     @form = form_for(@git_repository_location, edit_params)
 
     if @form.call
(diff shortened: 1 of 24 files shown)


sgerrand:
🙅 Please do not add this feature, for the following reasons:
* GitHub already supports linking external services, like CircleCI, against commits in a pull request
* GitHub already has branch protection, which prevents merges from occurring with configured status checks passing

This feature is duplicating a better implementation which is _both_ already available and being actively used in _every_ repository.

popovm:
@sgerrand You're correct, GitHub already have these checks and restrictions. 👍 
However, they might be omitted by developers with higher privileges, most of the time for a reason, of course. 
We could improve ST by keeping track of such out of the ordinary events. 
What if GitHub notifies ST that not all checks have passed? 💭

sgerrand:
> What if GitHub notifies ST that not all checks have passed? 💭

That's a good alternative. We can get a lot of information from the commit status.

sgerrand:
Note that Shipment Tracker already keeps track of commits which have failed CircleCI builds.

popovm:
Yes, as long as the CI builds(Jenkins or CircleCI) send information about test results to Shipment Tracker, it will save these results.
What it would not do now is triggering a deployment alert.

sgerrand:
> What it would not do now is triggering a deployment alert.

True, but should it? The act of deployment is what triggers evaluation of events which are attached to that commit, such as CircleCI test runs, JIRA ticket statuses, etc.

popovm:
I think Andy Steward would like to enter the discussion, as the idea of this feature came from him 🙂  
It is good if we are on the same page with him on that.

PR #223, opened 2017-09-28
[PO-3871] Clone repository if git folder is not present

mbdimitrova:
If the `.git` folder in the project directory is not present, the rugged repository will fail to get created. This will cause an error when a feature review for that repository is generated. To fix this the project will be cloned again when this error occurs and the rugged repository will be successfully created.

Relates:
- [x] JIRA :gem:: https://jira.fundingcircle.com/browse/PO-3871

/cc @FundingCircle/prod-ops 🙇‍♀️

### DIFF
--- a/app/models/git_repository_loader.rb
+++ b/app/models/git_repository_loader.rb
@@ -28,13 +28,7 @@ class GitRepositoryLoader
   def load(repository_name, update_repo: Rails.configuration.allow_git_fetch_on_request)
     git_repository_location = find_repo_location(repository_name)
 
-    repository = if update_repo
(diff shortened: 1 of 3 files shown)


app/models/git_repository_loader.rb
```diff
@@ -126,4 +119,14 @@ def instrument(name, &block)
       &block
     )
   end
+
+  def load_rugged_repository(update_repo:, git_repository_location:)
```
sgerrand: 🙋  Suggest changing the interface to this method slightly:
```ruby
def load_rugged_repository(update:, location:)
```
as the method already has `repository` as the context.

PR #268, opened 2018-01-03
Update moved tickets in Feature Reviews

mbdimitrova:
### What does this PR do?
It implements automatic update of the key of a ticket that has been moved in Jira. 

### Any background context you want to provide?
At the moment when a ticket is moved, the old ticket key remains in the feature reviews and cannot be linked because it cannot be found. This causes problem with the approval of feature reviews.

/cc @FundingCircle/prod-ops and @FundingCircle/engineering-effectiveness for review 🙇‍♀️

### DIFF
--- a/spec/factories/jira_event.rb
+++ b/spec/factories/jira_event.rb
@@ -159,6 +159,21 @@ FactoryGirl.define do
       )
       in_progress
     end
+
+    trait :moved do
+      changelog_details(
+        'changelog' => {
+          'items' => [
+            {
+              'field' => 'Key',
+              'fromString' => 'ONEJIRA-1',
+              'toString' => 'TWOJIRA-2',
+            },
+          ],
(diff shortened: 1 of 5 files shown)


spec/factories/jira_event.rb
```diff
@@ -159,6 +159,21 @@
...
     end
+
+    trait :moved do
+      changelog_details(
+        'changelog' => {
+          'items' => [
+            {
+              'field' => 'Key',
+              'fromString' => 'JIRA-1',
+              'toString' => 'JIRA-2',
```
sgerrand: 🙋 To further illustrate this functionality, consider naming these as two completely different project keys, e.g.
```diff
+              'fromString' => 'ONEJIRA-1',
+              'toString' => 'TWOJIRA-2',
```

PR #274, opened 2018-01-17
Add deployment alerts for devs that approve their own code

mbdimitrova:
### What does this PR do?
It adds an additional check for authorisation of tickets  - the user who has approved the 🎫  in Jira should be different from the user who worked on it 💻 

### Any specific implementation changes that would benefit highlighting?
* The ✉️  of the user that has moved the ticket in a **development** state should be different from the ✉️  of the user moving it in an **approved** state.
* Environment variable `DEVELOPMENT_STATUSES` should be set.

/cc @FundingCircle/prod-ops and @FundingCircle/engineering-effectiveness for review 🙇‍♀️

### DIFF
--- a/app/decorators/feature_review_with_statuses.rb
+++ b/app/decorators/feature_review_with_statuses.rb
@@ -78,7 +78,7 @@ class FeatureReviewWithStatuses < SimpleDelegator
   end
 
   def authorised?
-    @authorised ||= approved_by_owner? || (tickets.present? && tickets.all? { |t| t.authorised?(versions) })
+    @authorised ||= approved_by_owner? || (tickets.present? && tickets.all? { |t| t.authorised?(versions, isae_auditable?) })
   end
 
   def authorisation_status
@@ -127,5 +127,9 @@ class FeatureReviewWithStatuses < SimpleDelegator
 
   def git_repository_for(app_name)
     GitRepositoryLoader.from_rails_config.load(app_name)
+  end
+
+  def isae_auditable?
+    (@feature_review.app_names - Rails.application.config.isae_auditable).empty?
   end
 end
--- a/app/decorators/feature_review_with_statuses.rb
+++ b/app/decorators/feature_review_with_statuses.rb
@@ -78,7 +78,7 @@ class FeatureReviewWithStatuses < SimpleDelegator
   end
 
   def authorised?
-    @authorised ||= approved_by_owner? || (tickets.present? && tickets.all? { |t| t.authorised?(versions, isae_auditable?) })
+    @authorised ||= approved_by_owner? || (tickets.present? && tickets.all? { |t| t.authorised?(versions, isae_3402_auditable?) })
   end
 
   def authorisation_status
@@ -129,7 +129,7 @@ class FeatureReviewWithStatuses < SimpleDelegator
     GitRepositoryLoader.from_rails_config.load(app_name)
   end
 
-  def isae_auditable?
-    (@feature_review.app_names - Rails.application.config.isae_auditable).empty?
+  def isae_3402_auditable?
+    GitRepositoryLocation.where(name: @feature_review.app_names).all?(&:isae_3402_auditable?)
   end
 end
--- a/app/models/events/handlers/ticket_handler.rb
+++ b/app/models/events/handlers/ticket_handler.rb
@@ -13,7 +13,7 @@ module Events
           'summary' => event.summary,
           'status' => event.status,
           'event_created_at' => event.created_at,
-          'developed_by' => merge_developed_by(ticket, event),
+          'authored_by' => merge_authored_by(ticket, event),
           'approved_at' => merge_approved_at(ticket, event),
           'approved_by' => merge_approved_by(ticket, event),
         )
@@ -29,11 +29,11 @@ module Events
 
       private
 
-      def merge_developed_by(last_ticket, event)
+      def merge_authored_by(last_ticket, event)
         if event.development?
           event.user_email
         elsif last_ticket.present?
-          last_ticket['developed_by']
+          last_ticket['authored_by']
         end
       end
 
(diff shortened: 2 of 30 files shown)


app/models/ticket.rb
```diff
@@ -24,10 +26,14 @@ def approved?
...
-  def authorised?(versions_under_review)
-    return false if approved_at.nil?
+  def authorised?(versions_under_review, isae_auditable = false)
+    return false if approved_at.nil? || (isae_auditable && authorised_by_developer)
     linked_at = versions_under_review.map { |v| version_timestamps[v] }.compact.min
     return false if linked_at.nil?
     approved_at >= linked_at
   end
+
+  def authorised_by_developer
```
cassiomarques: Since this returns a boolean, should it be turned into a predicate method, with a `?` at the end?

spec/models/events/jira_event_spec.rb
```diff
@@ -60,6 +60,12 @@
     end
   end
 
+  describe '#development?' do
```
cassiomarques: Can we add a negative test case as well?

app/decorators/feature_review_with_statuses.rb
```diff
@@ -128,4 +128,8 @@ def commit_for_version(git_repository, version)
   def git_repository_for(app_name)
     GitRepositoryLoader.from_rails_config.load(app_name)
   end
+
+  def isae_auditable?
```
cassiomarques: Just curious, what does "ISAE" stand for?

bliof: It should be https://en.wikipedia.org/wiki/ISAE_3402

cassiomarques: @bliof Thanks!

sldblog:
From the last commit

> **Alert on approval by devs for certain repos only**
> Deployment alerts for tickets developed and approved by the same user should be applied for ISAE critical repos only.

It seems weird to track deployments in anything that is not auditable, as it seems to conflict with the purpose of this repository. The readme states:

> Tracks software versions during the release cycle for audit purposes.

Do we want to have anything here that is not explicitly for audit purposes?

mbdimitrova:
@sldblog @cassiomarques *ISAE* is the UK controls audit we are working towards and is required for big investors. `isae_auditable` refers to those repos that are ISAE critical and need change management, including preventing of code approval by the same user who developed it.

sldblog:
> [snip] refers to those repos that are ISAE critical and need change management, including preventing of code approval by the same user who developed it 

Should all repositories that are tracked with this application be considered ISAE critical?

If no, should we consider introducing different auditing/change management models (`:tracking_only`, `:isae_compliant`, etc.) to explicitly capture this intent?

sldblog:
😅 None of my comments are blocking, I'm just trying to understand.

mbdimitrova:
@sldblog No, @andysteward said that only a dozen of the applications tracked are considered ISAE critical. Instead of using the env variable, we can indeed introduce some kind of tracking setting for repositories. What would be the better solution?

sldblog:
@mbdimitrova Sounds like we should describe the level of audit necessary when registering a repository, next to the currently existing owners, approvers, URI settings. What do you think?

mbdimitrova:
@sldblog Sounds good! 👍 Will do it.

app/decorators/feature_review_with_statuses.rb
sgerrand: Please rename the method to `isae_3402_auditable?` to be explicit about what this method is referencing.

app/models/events/handlers/ticket_handler.rb
```diff
@@ -13,7 +13,9 @@ def apply
           'summary' => event.summary,
           'status' => event.status,
           'event_created_at' => event.created_at,
+          'developed_by' => merge_developed_by(ticket, event),
```
sgerrand: 🙋 Suggest renaming this to `authored_by`.

app/models/events/handlers/ticket_handler.rb
```diff
@@ -27,13 +29,29 @@ def feature_reviews_from_event
 
       private
 
+      def merge_developed_by(last_ticket, event)
```
sgerrand: 🙋 Also this to `merge_authored_by`.

config/application.rb
```diff
@@ -49,6 +49,10 @@ class Application < Rails::Application
     config.ssh_user = ENV['SSH_USER']
     config.approved_statuses = ENV.fetch('APPROVED_STATUSES', 'Ready for Deployment, Deployed, Done')
                                   .split(/\s*,\s*/)
+    config.development_statuses = ENV.fetch('DEVELOPMENT_STATUSES', 'In Progress, In Review')
+                                     .split(/\s*,\s*/)
+    config.isae_auditable = ENV.fetch('ISAE_AUDITABLE', '')
```
sgerrand: 🤔 Thinking about the tense here, should this be "auditable" or "audited"?

sgerrand:
Agree with @sldblog's comments here - I think that this extra level should be configurable per repository inside the application itself, rather than passed in as strings in an environment variable.

config/application.rb
mbdimitrova: 🤔 not sure, will think about it when refactoring.

mbdimitrova:
@sldblog @sgerrand @cassiomarques can you check the changes? 🙇‍♀️

app/decorators/feature_review_with_statuses.rb
```diff
@@ -128,4 +128,8 @@ def commit_for_version(git_repository, version)
   def git_repository_for(app_name)
     GitRepositoryLoader.from_rails_config.load(app_name)
   end
+
+  def isae_3402_auditable?
```
sgerrand: 💝

PR #284, opened 2018-02-21
Ruby 2.3.6 and circleci 2.0

bliof:
Update to circleci 2 and make sure that branches from dependabot are skipped as builds by default.


bliof:
I have not used alpine because for ruby 2.3.x there is some issue with the rugged gem / git `qsort_r`

.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
+version: 2
+
+jobs:
+  build:
+    branches:
+      ignore:
+        - /dependabot\/.*/
```
sgerrand: 🙋 Please don't ignore branches. It's OK if they fail.

.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+    docker:
+      - image: circleci/ruby:2.3
```
sgerrand: 🤔 Can you lock to the same version as in `.ruby-version`.

.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+      - image: circleci/postgres:9.4-alpine
+        environment:
+          POSTGRES_DB: shipment_tracker_test
+          POSTGRES_USER: st_test
+          POSTGRES_PASSWORD: ""
+    steps:
+      - run: sudo apt-get install -y cmake postgresql-client
+      - checkout
+      - run: cp .env.development.example .env.development
+      - run:  gem install bundler -v '1.16.1'
```
sgerrand: 🤔 Does the version _need_ to be specified? Does it matter?

.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+      - restore_cache:
+          key: shipment_tracker-gems-{{ checksum "Gemfile.lock" }}
+      - run: bundle check || bundle install
```
sgerrand: 💭 Might be worth specifying the `--path`. Not a blocker.

bliof: It is configured with env var

.circleci/config.yml
bliof: This is mainly because of the

```
BUNDLED WITH
   1.16.1
```

lines in the `Gemfile.lock`.

I forced it to hit the cache. 🤔 `gem install bundler` should be enough

.circleci/config.yml
bliof: My problem is that they are a lot ~35 and for each merge the bot will rebase and schedule 35 jobs. Let's keep it like this so we can go through the PRs faster.

TBO I would prefer if it can add a [skip ci] in the commit message, so when we do changes to the PR they'll trigger the ci.

ps. I think like this it will create the skipped build in circle ci and you can click rebuild from the web ui.

sgerrand: They're a lot at the moment, but if we merge them more frequently then they'll drop in number.

.circleci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+          SHIPMENT_TRACKER_DATABASE_PASSWORD: ""
+      - image: circleci/postgres:9.4-alpine
+        environment:
+          POSTGRES_DB: shipment_tracker_test
+          POSTGRES_USER: st_test
+          POSTGRES_PASSWORD: ""
+    steps:
+      - run: sudo apt-get install -y cmake postgresql-client
+      - checkout
+      - run: cp .env.development.example .env.development
```
sgerrand: 🙋 Shouldn't this be `cp .env.development.example .env.test`?

bliof: one sec 🤔

PR #287, opened 2018-02-22
Update GitHub status only when changed

mbdimitrova:
In order to reduce updates sent to GitHub, only post status when the new status is different from the previous. For more information see:
https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref

@FundingCircle/prod-ops, @FundingCircle/engineering-effectiveness 🙇‍♀️

### DIFF
--- a/features/support/scenario_context.rb
+++ b/features/support/scenario_context.rb
@@ -110,9 +110,12 @@ module Support
         comment_body: comment_type.build_comment(url),
         updated: time,
       )
+
       @stubbed_requests['pending'] = stub_request(:post, %r{https://api.github.com/.*})
                                      .with(body: /"state":"pending"/)
                                      .and_return(status: 201)
+      stub_github_status
+
       event = build(:jira_event, ticket_details)
       travel_to Time.zone.parse(time) do
         post_event 'jira', event.details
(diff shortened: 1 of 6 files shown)


features/support/scenario_context.rb
```diff
@@ -146,6 +149,11 @@ def stub_github_update_for_repo_owner(action)
                                  .and_return(status: 201)
     end
 
+    def stub_github_status
```
sgerrand: 🙋 Suggest either renaming this method to `stub_github_status_as_successful` or alter the signature to allow the status code to be passed in, e.g.
```ruby
def stub_github_status(status_code:)
```
Otherwise the method name is somewhat ambiguous.

lib/clients/github.rb
```diff
@@ -30,6 +30,15 @@ def repo_accessible?(uri)
     false
   end
 
+  def last_status_for(repo:, sha:)
+    client.combined_status(repo, sha)[:statuses].reverse.find do |status|
+      status[:context] == 'shipment-tracker'
+    end
+  rescue Octokit::Error
```
sgerrand: 🙋 `Octokit::Error` can represent a [lot of different error types](https://github.com/octokit/octokit.rb/blob/master/lib/octokit/error.rb#L16-L31). Consider logging the error `type`, if not the `message` as well.

PR #289, opened 2018-03-02
Fix Github status updates

mbdimitrova:
Change the condintion when a status should be reported


sgerrand:
@mbdimitrova or @bliof: Please update the description for this issue with details on exactly what the bug was and how this change remediates the issue.

bliof:
@sgerrand refresh your browser :trollface:

sgerrand:
That's great, thanks!

PR #295, opened 2018-03-16
Update the repo in the relink_ticket_job

bliof:
A lot of times the repo is no updated when running this job and commit
new commit sha is not found. This causes and exception and the task
fails..

### DIFF
--- a/app/jobs/relink_ticket_job.rb
+++ b/app/jobs/relink_ticket_job.rb
@@ -58,7 +58,7 @@ class RelinkTicketJob < ActiveJob::Base
   end
 
   def commit_on_master?(full_repo_name, sha)
-    git_repo = GitRepositoryLoader.from_rails_config.load(full_repo_name.split('/')[1])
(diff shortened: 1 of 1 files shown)


app/jobs/relink_ticket_job.rb
```diff
@@ -58,7 +58,7 @@ def post_error_status(full_repo_name, sha)
   end
 
   def commit_on_master?(full_repo_name, sha)
```
0mega: I'm not 💯 sure, but there's a chance that this method is used in UI, and it can take quite long time to update some repos which might render a timeout for users (because of default 60 sec. http timeout). I think we had this problem before that's why I'm letting you know.
Just saying this so that you double check please 🙏

0mega: 🤦‍♂️ it's a job, forget my comment

bliof: There is one in GitRepository ;)

The current one is only for the job.

PR #297, opened 2018-03-16
Update ticket's authored_by with assignee email if needed 

mbdimitrova:
When handling a Jira approval event and creating a ticket, if the previous ticket with that Jira key doesn't have authored_by, set it to the event's assignee email 💭 

This will fix the following bug: when a ticket is moved in a _development_ status, then linked in Shipment Tracker and then approved ➡️ authored_by is `nil` which might lead to _false positives_.

/cc @FundingCircle/engineering-effectiveness, @FundingCircle/prod-ops for review 🙇‍♀️

### DIFF
--- a/spec/models/events/handlers/ticket_handler_spec.rb
+++ b/spec/models/events/handlers/ticket_handler_spec.rb
@@ -7,6 +7,8 @@ RSpec.describe Events::Handlers::TicketHandler do
     let(:ticket_defaults) { { paths: [path], versions: %w(foo), version_timestamps: { 'foo' => nil } } }
     let(:url) { feature_review_url(app: 'foo') }
     let(:path) { feature_review_path(app: 'foo') }
+    let(:email) { 'test@example.com' }
+    let(:email2) { 'test2@example.com' }
 
     it 'updates key information' do
       ticket = {}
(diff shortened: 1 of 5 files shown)


spec/models/events/jira_event_spec.rb
```diff
@@ -95,4 +95,10 @@
       expect(build(:jira_event, :created).user_email).to eq 'joe.bloggs@example.com'
     end
   end
+
+  describe '#assignee_email' do
+    it 'return the email of the assignee' do
```
sgerrand: 🙋 "returns"

spec/models/events/handlers/ticket_handler_spec.rb
```diff
@@ -32,6 +34,35 @@
 
         expect(new_ticket).to include('approved_at' => time + 1.hour)
       end
+
+      it "updates authored_by with event's assignee email if the previous ticket does not have authored_by" do
```
sgerrand: Suggest reordering this slightly to something like the following
```ruby
context "if the previous ticket" do
  context "is missing a value for authored_by" do
    it "updates authored_by with event's assignee email"
  end
  context "is missing a value for authored_by" do
    it "updates authored_by with previous value of authored_by"
  end
end
```
That will enable using `let`s to reduce the logic to a single shared example. Not a blocker though.

spec/models/events/handlers/ticket_handler_spec.rb
```diff
@@ -32,6 +34,35 @@
...
+        ticket = {}
+        created_ticket_event = build(:jira_event, :approved, created_at: time + 1.hour, assignee_email: email)
+
+        new_ticket = described_class.new(ticket, created_ticket_event).apply
+
+        expect(new_ticket).to include('authored_by' => email)
+      end
+
+      it "updates authored_by with previous ticket\'s authored_by if the previous ticket has authored_by" do
```
bliof: 👀  `\'`

PR #334, opened 2018-08-07
Mesos

maciekm:
The changes update the application to facilitate running as a Docker container on Mesos.

Keeps the same service names as on Rackspace.
`git-worker`
`event-worker`
`delayed-job`

#### Notable changes
- All containers have the local git cache mounted to `/efs/shipment-tracker` on the Mesos slave. This currently requires running the containers as `root` instead of the unprivileged `shipment_tracker` user. Most of the apps running on Mesos are doing that so 🤷‍♂️. To enable all containers to be reasonable in sync with their view of EFS contents, the `actimeo` mount option was set to 3 https://github.com/FundingCircle/ayn/pull/2056.
- The git repositories are cloned with the [`bare`](https://www.rubydoc.info/github/libgit2/rugged/Rugged%2FRepository%2Ebare.) option to improve performance and make it more obvious that the working directory contents are not used at all.
- Uses SES for deploy alert emails to repository owners.
- `git-worker` and `event-worker` Rake tasks terminate gracefully on `SIGINT`/`SIGTERM`.
- Fixes gem deprecation warnings.
- Removes request timeout configuration that was set for Heroku.
- Adds a configurable `LOG_LEVEL`
- Removes `database.yml` and uses `DATABASE_URL` instead.
- Adds more logging.
- Uses up to 16 threads by default in `git-worker` task, instead of 4. This is now configurable.


docker/dropzone.yaml
```diff
@@ -0,0 +1,48 @@
+github:
+  project_name: shipment_tracker 
+deploy_tasks:
+  lock_flight:
+    enabled: true
+  shipment_tracker:
+    enabled: true
+    deploy_token: KxT3S6cFjtQ5aVsB3GSVrzQW
```
fussybeaver: is this a secret ? If so, with dropzone it's possible to inject these through environment variables

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM quay.io/fundingcircle/alpine-ruby:2.3 as builder
+
+RUN apk --no-cache add linux-headers postgresql-dev build-base cmake libgit2-dev 
+WORKDIR /tmp
+COPY .ruby-version /tmp/
+COPY Gemfile* /tmp/
+RUN bundle config build.libv8 --with-system-v8 && \
```
fussybeaver: why does shipment tracker need the chrome bindings?

docker/dropzone.yaml
sgerrand: It's not that secret: https://github.com/search?q=org%3AFundingCircle+deploy_token&type=Code

Dockerfile
sgerrand: It's needed for the `therubyracer` gem used by the Rails asset pipeline.

sgerrand: That said, @maciekm – you could probably use `nodejs` instead and remove `therubyracer` altogether. 🤷‍♀️

PR #350, opened 2018-11-07
Dockerize the app for local development

LS80:
#### Changes
- [x] Introduces Docker Compose to run the app locally. Some tweaks were required to fix `rugged` with git clone over ssh.
- [x] Cleans up the logs by fixing various warning messages.
- [x] Makes the git repository directory configurable.

#### Notes
Production uses Postgres 9.4 so this installs the older client and libs from Alpine 3.3.

I just discovered https://github.com/FundingCircle/shipment_tracker/pull/334 by @maciekm for Mesos. Hopefully that can be rebased on this. There are some differences that we can discuss such as using `nodejs` or `therubyracer`.

#### Secrets required to be set in environment
- `AUTH0_CLIENT_SECRET` from https://manage.auth0.com/#/applications/Y9ltgcdXUkNA4hRDWuP0nxDaZVdIskBM/settings.
- `GITHUB_REPO_READ_TOKEN` from https://github.com/settings/tokens/new.
- `SSH_PUBLIC_KEY` and `SSH_PRIVATE_KEY` without passphrase and added to GitHub account.

image

image

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,13 +3,12 @@ FROM quay.io/fundingcircle/alpine-ruby:2.3 as builder
 RUN apk --no-cache add \
   cmake \
   linux-headers \
+  postgresql-dev \
   sqlite-dev \
   zlib-dev \
  && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.8/main \
   libgit2-dev \
-  libssh2-dev \
- && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.3/main \
-  postgresql-dev==9.4.15-r0
+  libssh2-dev
 
 WORKDIR /tmp
 COPY Gemfile* ./
@@ -21,17 +20,16 @@ LABEL maintainer="Funding Circle Engineering <engineering@fundingcircle.com>"
 
 RUN apk --no-cache add \
   nodejs \
+  postgresql \
+  postgresql-client \
   sqlite-dev \
   tzdata \
   zlib \
  && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.8/main \
   libgit2 \
-  libssh2 \
- && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.3/main \
-  libpq==9.4.15-r0 \
-  postgresql-client==9.4.15-r0 \
-  postgresql==9.4.15-r0 \
- && addgroup -g 1101 -S shipment_tracker && \
+  libssh2
+
+RUN addgroup -g 1101 -S shipment_tracker && \
   adduser -S -u 1101 -h /app -G shipment_tracker shipment_tracker
 
 USER shipment_tracker
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,38 @@
+FROM quay.io/fundingcircle/alpine-ruby:2.3 as builder
+
+RUN apk --no-cache add \
+  cmake \
+  linux-headers \
+  sqlite-dev \
+  zlib-dev \
+ && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.3/main \
+  postgresql-dev==9.4.15-r0
+
+WORKDIR /tmp
+COPY Gemfile* ./
+RUN bundle install --deployment --without deployment dockerignore
+
+FROM quay.io/fundingcircle/alpine-ruby:2.3
+LABEL maintainer="Funding Circle Engineering <engineering@fundingcircle.com>"
+
+RUN apk --no-cache add \
+  nodejs \
+  sqlite-dev \
+  tzdata \
+  zlib \
+ && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.3/main \
+  libpq==9.4.15-r0 \
+  postgresql-client==9.4.15-r0 \
+  postgresql==9.4.15-r0 \
+ && addgroup -g 1101 -S shipment_tracker && \
+  adduser -S -u 1101 -h /app -G shipment_tracker shipment_tracker
+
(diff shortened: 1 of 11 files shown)


cddr:
I wonder if the configuration would be simplified with something like this. Then if you need to create a new key specifically for this purpose (which seems like a good idea from a security point of view), you just run `ssh-keygen -f id_rsa.shipment-tracker` in the current directory and upload the resulting key to github. Removes two out of the four environment variable that need to be configured by the user.

```
diff --git a/.gitignore b/.gitignore
index a30cdf0..4fe9a43 100644
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@
 /config/database.yml

 .rspec
+
+id_rsa.shipment-tracker
+id_rsa.shipment-tracker.pub
\ No newline at end of file
diff --git a/docker-compose.yml b/docker-compose.yml
index afdfd74..1366426 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,8 +19,8 @@ services:
       SHIPMENT_TRACKER_DATABASE_PASSWORD: postgres
       SHIPMENT_TRACKER_DATABASE_USERNAME: postgres
       SSH_USER: git
-      SSH_PUBLIC_KEY:
-      SSH_PRIVATE_KEY:
+      SSH_PUBLIC_KEY: id_rsa.shipment-tracker.pub
+      SSH_PRIVATE_KEY: id_rsa.shipment-tracker
     ports:
       - 127.0.0.1:1201:1201
     depends_on:
```

LS80:
We can look at doing that in a follow up PR 👍
It would also be good to allow the use of an SSH key with a passphrase.

docker-compose.yml
```diff
@@ -0,0 +1,57 @@
+version: '2.4'
+services:
+  web:
+    build:
+      context: .
+    image: shipment_tracker
+    command: sh -c 'bundle exec rake db:create db:migrate && exec bundle exec rails server --binding=0.0.0.0 --port 1201'
+    environment:
```
sgerrand:
```
$ docker-compose run web
Creating network "shipment_tracker_default" with the default driver
Creating volume "shipment_tracker_pgdata" with default driver
Creating volume "shipment_tracker_git" with default driver
Creating shipment_tracker_postgres_1 ... done
could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 5432?
could not connect to server: Address not available
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 5432?
```

sgerrand: Suggest adding
```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index afdfd74..5a4ee21 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,7 @@ services:
       AUTH0_CLIENT_SECRET:
       AUTH0_CONNECTION: fundingcircle.com
       AUTH0_DOMAIN: fundingcircle.auth0.com
+      DATABASE_URL: postgres://postgres:5432/shipment_tracker
       GITHUB_REPO_READ_TOKEN:
       GIT_REPOSITORY_CACHE_DIR: /tmp/git
       PROTECT_STDOUT: 'true'
```
to fix the issue.

docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+      - 127.0.0.1:1201:1201
+    depends_on:
+      - postgres
+    volumes:
+      - git:/tmp/git
+
+  update_git_loop:
+    image: shipment_tracker
+    command: sh -c 'bundle exec rake db:create db:migrate && exec bundle exec rake jobs:update_git_loop'
+    environment:
```
sgerrand: Also needed here.

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+RUN apk --no-cache add \
+  nodejs \
+  sqlite-dev \
+  tzdata \
+  zlib \
+ && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.8/main \
+  libgit2 \
+  libssh2 \
+ && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.3/main \
```
sgerrand: 🙋 The mix of Alpine Linux versions (3.3 and 3.8) is not particularly great. Is there a specific need for using this older version of the PostgreSQL libraries? They're normally backwards compatible.

Gemfile
```diff
@@ -25,12 +25,15 @@ gem 'rugged', '~> 0.27.0'
 gem 'sass-rails'
 gem 'slack-notifier'
 gem 'solid_use_case'
-gem 'therubyracer'
 gem 'uglifier'
 gem 'unicorn-rails'
 gem 'unicorn'
 gem 'virtus'
 
+group :dockerignore do
```
sgerrand: 👏

docker-compose.yml
LS80: I can't reproduce the issue. The database configuration comes from the `SHIPMENT_TRACKER_DATABASE_...` environment variables.

Dockerfile
LS80: As you say, we can probably get away with using the Alpine 3.5 bundled version of the Postgres libraries. However, I had to use the latest `libgit2-dev` from Alpine 3.8 because `rugged` links to `libgit2.so.27`, which is not provided by earlier versions. I also had to use the latest `libssh2` for the repositories to clone successfully. I haven't checked the reason for that one yet.

We are currently tied to Alpine 3.5 by using Ruby 2.3. If you are OK with updating to Ruby 2.5 we can use Alpine 3.8 for everything.

sgerrand: Happy with pulling in `libgit2` and `libssh` from Alpine 3.8. We (EE) can do the Ruby upgrade separately. 🙇

PR #377, opened 2019-01-30
Prevent unnecessary repo re-clone caused by concurrent fetches

LS80:
The git repositories are updated by both the `git-worker` process and the `RelinkTicketJob` delayed-job.

`libgit2` creates a lock file when it updates a reference in a git fetch operation, e.g. `/tmp/6-funding_circle_app/.git/refs/remotes/origin/di-ri-928.lock`. If this fetch operation is in progress when a `libgit2` fetch from the other process tries to update the same reference, it causes an exception
```
Rugged::OSError: failed to create locked file '/tmp/6-funding_circle_app/.git/refs/remotes/origin/di-ri-928.lock': File exists 
```
We currently re-clone the repository when there is any `Rugged::OSError`, but this particular one is recoverable without doing a clone, which can take a long time for a large repository.

This will instead retry a git fetch up to 10 times if a concurrent git fetch has a lock on the same reference. It sleeps for 1 second before each retry.

Fixes e.g. https://app.honeybadger.io/projects/44071/faults/38110914/dc7c0778-1fea-11e9-8926-2faa98b2053c#notice-trace

See https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/5497

### DIFF
--- a/app/models/git_repository_loader.rb
+++ b/app/models/git_repository_loader.rb
@@ -10,6 +10,8 @@ class GitRepositoryLoader
   class NotFound < RuntimeError; end
   class BadLocation < RuntimeError; end
 
+  RETRY_LIMIT = 10
+
   def self.from_rails_config
     @repo_loader ||= new(
       ssh_private_key: Rails.configuration.ssh_private_key,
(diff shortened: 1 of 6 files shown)


app/models/git_repository_loader.rb
```diff
@@ -49,13 +51,15 @@ def find_repo_location(repository_name)
 
   def updated_rugged_repository(git_repository_location, options)
     dir = repository_dir_name(git_repository_location)
-    Rugged::Repository.new(dir, options).tap do |repository|
+    @retries ||= 0
```
bliof: 🤔 shouldn't this be 

```
@retries = 0
begin
```

and how about using a local variable that is passed around `retryable_fetch?(error, retries)`?

bliof: actually you don't have to pass it around, just don't check it in `retryable_fetch`

LS80: Which method are you suggesting to drop?

bliof: the `retryable_fetch?`

e.g. something like

```
retries = 0

begin
  repo.fetch('origin', options) unless up_to_date?(git_repository_location, repo)
rescue Rugged::OSError, Rugged::RepositoryError, Rugged::InvalidError, Rugged::ReferenceError => error
  if retryable_fetch?(error) && retries <= RETRY_LIMIT
    retries += 1
    retry
  end
end
```

LS80: I started with pretty much exactly that, but RuboCop.

LS80: Method too long, assignment branch condition, yada yada...

bliof: you can move the fetch and the clone into separate methods 🙈 

may only issue is that the current retry is global for the class e.g. it will be weird to iterate over each repository and do an update using one instance of the loader

LS80: Good point, the retry count should be separate for each repository. I'll fix it 👍

PR #385, opened 2019-02-12
Add a Delayed Job web interface

LS80:
Uses https://github.com/ejschmitt/delayed_job_web to enable better visibility of number of enqueued jobs, failed jobs etc.

Will be at https://shipment-tracker.fundingcircle.co.uk/delayed_job/overview

Secured via Auth0 (session expires after 24 hours).


LS80:
It is not secured. I'll see how easy it is to add it through `omniauth-auth0`.

config/initializers/omniauth.rb
```diff
@@ -5,17 +5,22 @@
 Rails.configuration.login_url          = '/auth/auth0'
 Rails.configuration.login_callback_url = '/auth/auth0/callback'
 
+auth0_provider_args = {
+  client_id:     ENV.fetch('AUTH0_CLIENT_ID', nil),
+  client_secret: ENV.fetch('AUTH0_CLIENT_SECRET', nil),
+  namespace:     ENV.fetch('AUTH0_DOMAIN', 'fundingcircle.auth0.com'),
```
sgerrand: 🙋 Please don't set a default value for this environment variable.

config/initializers/omniauth.rb
```diff
@@ -5,17 +5,22 @@
...
+  connection:    ENV.fetch('AUTH0_CONNECTION', 'Username-Password-Authentication'),
```
sgerrand: Suggest not setting a default for this either.

config/initializers/omniauth.rb
LS80: I didn't change any of the defaults. The `omniauth` configuration is the same as it was.

sgerrand: 🙇 Will resolve separately.

PR #393, opened 2019-02-20
Ruby 2.3.8

sgerrand:
💁 Updates the Ruby version to 2.3.8, the [latest generally available version](https://www.ruby-lang.org/en/downloads/) in the 2.3.x stream.


LS80:
The Ruby version is 2.5 since https://github.com/FundingCircle/shipment_tracker/pull/411.

PR #394, opened 2019-02-20
Send a deploy alert email to the person deploying

LS80:
Currently deploy alert emails are sent only to the repository owners. Alerts are also sent to the #deployment-alerts on Slack but not everyone monitors that channel.


sgerrand:
🤔 I'm somewhat concerned that the recipients are unlikely to read the messages that are generated.

LS80:
Yeah quite possibly. Slack deploy alerts also regularly go unnoticed by the deployer. This is an attempt to make it more likely that immediate action can be taken by the deployer without needing to be prompted by others.

PR #412, opened 2019-04-15
Auto-link feature reviews to JIRA tickets when GH branch name contains ticket key

kylejm:
## Summary
This PR enables Shipment Tracker to automatically link Feature Reviews to JIRA tickets when the GitHub branch for the given feature contains the JIRA ticket key, which is common practice for my team and for at least a few other teams at Funding Circle.

## Motivation
My team and I were finding linking tickets to feature reviews tedious. We realised since we always put the ticket key in our branch names (`km-some-feature-MOB-123`), we're already creating a relationship between the branch/feature and a JIRA ticket, so it would be great if Shipment Tracker recognised that.

## How It Works
The idea is that Shipment Tracker will check the GitHub branch using a regex to search of a substring containing characters in the following order:
* 2-8 capital letters
* A hyphen
* One or more digits starting with a digit between 1 and 9

For example: MOB-123, FCA-345, MIGRATE-5604.

This means there is possibility for auto-link to incorrectly link a ticket if the developer happens to name a branch with capitals, a hyphen and a number that make up a valid ticket key in our JIRA instance, however I feel this is quite unilkey. A valid sequence of the above mentioned characters is likely to be the key of the ticket that the developer is working on.

Alternatively, we could make auto-linking safer by instructing developers to enclose the ticket key they're working on with some special characters if they want Shipment Tracker to auto-link.

## Comments/Discussion
I'm nowhere near a Ruby/Rails pro, and am not very familiar with best practices or styles. There are some questionable areas I have about the code I've written so far. I'm also trying to get a feel for Shipment Trackers code and understand the correct approach for auto-link as a whole.

Below are some things I'd like help with.

### Regex
* Any feedback on the regex used?
* I'm pretty sure this should be a constant somewhere? 
* And it should be frozen?

### Returning a Bool from `check_branch_for_ticket_and_link?`
The idea of `check_branch_for_ticket_and_link?` returning a Bool feels a bit odd. We need a way of the `perform` function understanding if auto-linking processed as expected or if there was an error. I'd like to discuss this approach. 

I noticed that we were using `@send_error_status`, but this current approach seems flawed. Consider the current code on `master` for this:
```ruby
     if branch_created || commit_on_master?(full_repo_name, after_sha) ||
       relink_tickets(before_sha, after_sha).empty?
      post_not_found_status(full_repo_name, after_sha)
    elsif @send_error_status
      post_error_status(full_repo_name, after_sha)
    end
```
Notice that `post_error_status(full_repo_name, after_sha)` is only run if `relink_tickets` returns empty **and** `@send_error_status` is set to true, which is not possible? So it feels like error handling might need a bigger rethink all together.

### Building the Feature Release URL
I'm struggling to understand how to build a URL to pass to `LinkTicket`. I figured it's just the defined route in `routes.rb` for `:feature_reviews`, but want to be sure that using the commented code I've left for this is the right idea to go with.

### Anything Else
Please give me any other pointers on my overall approach and idea. After gathering some feedback I'll prepare some tests and finish off the implementation.

---

Thanks for taking the time to read! ❤️


app/jobs/relink_ticket_job.rb
```diff
@@ -36,9 +44,22 @@ def relink_tickets(before_sha, after_sha)
     end
   end
 
+  def check_branch_for_ticket_and_link?(branch_name, after_sha)
+    /(?<ticket_key>[A-Z]{2,8}-[1-9][0-9]*)/ =~ branch_name
+    return true if ticket_key.nil?
+
+    url = '' # TODO: build url - "#{Rails.application.routes.url_helpers.feature_reviews}/#{app_name}?#{params.to_query}" ?
```
0mega: You can check how to build the url from [here](https://github.com/FundingCircle/shipment_tracker/blob/001b012cb7bee329002ee8ed7f9974b98b150182/app/models/commit_status.rb#L82)

app/jobs/relink_ticket_job.rb
```diff
@@ -36,9 +44,22 @@ def relink_tickets(before_sha, after_sha)
     end
   end
 
+  def check_branch_for_ticket_and_link?(branch_name, after_sha)
+    /(?<ticket_key>[A-Z]{2,8}-[1-9][0-9]*)/ =~ branch_name
```
0mega: You are correct, this should be a frozen constant, similar to [this](https://github.com/FundingCircle/shipment_tracker/blob/1e616cb6cff91bb2dbf754019f52a1a87344463d/app/models/forms/repository_locations_form.rb#L19).

app/jobs/update_ticket_links_job.rb
```diff
@@ -59,9 +80,12 @@ def post_error_status(full_repo_name, sha)
...
-  def commit_on_master?(full_repo_name, sha)
-    git_repo = GitRepositoryLoader.from_rails_config.load(full_repo_name.split('/')[1], update_repo: true)
+  def extract_ticket_key_from_branch_name(branch_name)
+    /(?<ticket_key>[A-Z]{2,8}-[1-9][0-9]*)/ =~ branch_name
+    ticket_key
+  end
 
-    git_repo.commit_on_master?(sha)
+  def url_for_repo_and_sha(full_repo_name, sha)
+    CommitStatus.new(full_repo_name: full_repo_name, sha: sha).target_url
```
kylejm: @sgerrand I took a look at improving the URL creation, but struggled a bit as we do it in so many different ways across this codebase.

This is the solution I came to in the end, but it's still a WIP as this only provides the path and not the host, therefore we'd need a way of feeding the host into the job. Tbf, explicitly handing the host to the job probably makes sense, as it's probably not safe to use the job executors host to create url?

```suggestion
    repo_name = full_repo_name.split('/').last
    Factories::FeatureReviewFactory.new.create_from_apps({ repo_name => sha }).path
```

Hope that makes sense, catch me before 5:30pm if it doesn't. Otherwise, do whatever you feel fit whilst I'm away.

spec/jobs/update_ticket_links_job_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      let(:args) do
+        {
+          full_repo_name: full_repo_name,
+          before_sha: before_sha,
+          after_sha: after_sha,
+          branch_created: true,
+          branch_name: 'master'
+        }
+
+        it 'should not link any tickets' do
```
0mega: might be a good idea to replace all "should" occurrences in this file with 3rd person to follow the convention in the rest of the project. Like "does not link any tickets". Found "should" being used in 3 specs in the whole project 🙈.

app/use_cases/handle_push_event.rb
```diff
@@ -37,7 +37,7 @@ def relink_tickets(payload)
       before_sha: payload.before_sha,
       after_sha: payload.after_sha,
       branch_created: payload.branch_created?,
-      branch_name: payload.branch_name
+      branch_name: payload.branch_name,
```
filipebarros: this is not needed

spec/jobs/update_ticket_links_job_spec.rb
```diff
@@ -46,15 +46,15 @@
...
         }
       end
 
       it 'should link the correct ticket key' do
         expected_comment = "[Feature ready for review|#{expected_url}]"
         expect(JiraClient).to receive(:post_comment).with(ticket_key, expected_comment)
-        stub = stub_request(:post, "https://api.github.com/repos/#{full_repo_name}/statuses/#{after_sha}").
-         to_return(status: 200)
+        stub = stub_request(:post, "https://api.github.com/repos/#{full_repo_name}/statuses/#{after_sha}")
+               .to_return(status: 200)
```
filipebarros: 👏

app/jobs/update_ticket_links_job.rb
```diff
@@ -59,9 +84,12 @@ def post_error_status(full_repo_name, sha)
     CommitStatus.new(full_repo_name: full_repo_name, sha: sha).error
   end
 
-  def commit_on_master?(full_repo_name, sha)
-    git_repo = GitRepositoryLoader.from_rails_config.load(full_repo_name.split('/')[1], update_repo: true)
+  def extract_ticket_key_from_branch_name(branch_name)
+    /(?<ticket_key>[A-Z]{2,8}-[1-9][0-9]*)/ =~ branch_name
```
LS80: Our project key limit is actually 10 characters, which is the default. As it's configurable in the Jira instance we might want to increase it or make it configurable. For now making it 10 instead of 8 should be fine?

Make it case insensitive too?
```suggestion
    /(?<ticket_key>[A-Z]{2,10}-[1-9][0-9]*)/i =~ branch_name
```

spec/jobs/update_ticket_links_job_spec.rb
kylejm: Updated 👍

app/jobs/relink_ticket_job.rb
kylejm: Worked it out 👍

app/jobs/relink_ticket_job.rb
kylejm: Done 👍

app/jobs/update_ticket_links_job.rb
kylejm: Updated to 10 👍.

I was thinking that making it case sensitive would give devs a way to be clearer that a ticket key is indeed a ticket key, since if it's lower case allowed the following branch would be seen as containing a ticket: `some-feature-123`. What do you think?

LS80: Yep makes sense, upper case is fine 👍

app/jobs/update_ticket_links_job.rb
```diff
@@ -31,14 +37,35 @@ def relink_tickets(before_sha, after_sha)
...
+  def check_and_link_new_branch(full_repo_name, branch_name, after_sha)
+    if check_branch_for_ticket_and_link?(full_repo_name, branch_name, after_sha)
+      post_not_found_status(full_repo_name, after_sha)
+    else
+      post_error_status(full_repo_name, after_sha)
+    end
+  end
+
+  def check_branch_for_ticket_and_link?(full_repo_name, branch_name, after_sha)
+    ticket_key = extract_ticket_key_from_branch_name branch_name
```
0mega: Just to keep the style consistent
```suggestion
    ticket_key = extract_ticket_key_from_branch_name(branch_name)
```

app/jobs/update_ticket_links_job.rb
```diff
@@ -23,6 +27,8 @@ def perform(args)
 
   private
 
+  JIRA_TICKET_REGEX = /(?<ticket_key>[A-Z]{2,10}-[1-9][0-9]*)/.freeze
```
0mega: Constants usually defined at the top (at least in this proj. if I remember prop.). Also if you want to make it private this wouldn't work, you need something like [this](http://www.chrisrolle.com/en/blog/private-constants-in-ruby)

app/jobs/update_ticket_links_job.rb
```diff
@@ -31,14 +37,35 @@ def relink_tickets(before_sha, after_sha)
...
+    end
+  end
+
+  def check_branch_for_ticket_and_link?(full_repo_name, branch_name, after_sha)
+    ticket_key = extract_ticket_key_from_branch_name branch_name
+
+    return true if ticket_key.nil?
+
+    link_feature_review_to_ticket(ticket_key, url_for_repo_and_sha(full_repo_name, after_sha))
+    !@send_error_status
```
0mega: This is a little bit confusing, as in, what's the role it's playing here, also looks like it might be nil, cause it's only defined if `JiraClient.post_comment(ticket_key, LinkTicket.build_comment(url))` throws an exception, in that case it might not behave as you want it too (I think).
You could initialize it to `false` in this method.

app/models/payloads/github.rb
```diff
@@ -42,6 +42,10 @@ def branch_deleted?
       @data['deleted']
     end
 
+    def branch_name
+      @data['ref'].split('/')[2]
```
0mega: Would this work if I create a branch like `feature/my-super-feature-TICKET-123`? As in, would it still be index 2 or do we just want to take the path after the last `/`?

app/models/payloads/github.rb
```diff
@@ -42,6 +42,10 @@ def branch_deleted?
       @data['deleted']
     end
 
+    def branch_name
```
0mega: Would be nice to add tests for it [here](https://github.com/FundingCircle/shipment_tracker/blob/master/spec/models/payloads/github_spec.rb)

app/models/payloads/github.rb
kylejm: Yes, the branch name is always the last section of three parts divided by `/`. It goes `refs/REF_TYPE/REF_NAME`.

app/jobs/update_ticket_links_job.rb
kylejm: Done 👍

app/jobs/update_ticket_links_job.rb
kylejm: Good catch, thank you. Updated to make it more readable.

app/jobs/update_ticket_links_job.rb
```diff
@@ -31,14 +37,27 @@ def relink_tickets(before_sha, after_sha)
...
+        update_path_and_link_feature_review_to_ticket(ticket.key, feature_review_path, before_sha, after_sha)
       end
     end
   end
 
-  def link_feature_review_to_ticket(ticket_key, old_feature_review_path, before_sha, after_sha)
+  def check_branch_nane_for_ticket_and_link(full_repo_name, branch_name, after_sha)
+    ticket_key = extract_ticket_key_from_branch_name(branch_name)
+
+    post_not_found_status(full_repo_name, after_sha) && return unless ticket_key
```
LS80: Could also be
```suggestion
    return post_not_found_status(full_repo_name, after_sha) unless ticket_key
```

kylejm: Is this usual for ruby? I'm a n00b so this might not be a valid view, but `return post_not_found_status` makes me thing that `post_not_found_status` returns something that our method is also interested in returning, which it's not.

app/models/payloads/github.rb
kylejm: Done 👍

app/models/payloads/github.rb
kylejm: Updated to use a regex for extracting the branch name. Good catch Olek on the case where a branch name contains `/` 👏 🙇

kylejm:
Sorry for all the pushes everyone, I never got spec running on local because of some dep setup that I didn't understand.

app/jobs/update_ticket_links_job.rb
LS80: Depends what behaviour you want. As it stands if `post_not_found_status` returns `nil` then there will not be an early return.

app/jobs/update_ticket_links_job.rb
```diff
@@ -31,14 +37,27 @@ def relink_tickets(before_sha, after_sha)
...
         next unless feature_review_path.include?(before_sha)
 
-        link_feature_review_to_ticket(ticket.key, feature_review_path, before_sha, after_sha)
+        update_path_and_link_feature_review_to_ticket(ticket.key, feature_review_path, before_sha, after_sha)
       end
     end
   end
 
-  def link_feature_review_to_ticket(ticket_key, old_feature_review_path, before_sha, after_sha)
+  def check_branch_nane_for_ticket_and_link(full_repo_name, branch_name, after_sha)
```
LS80: Just noticed typo `nane` -> `name`

app/jobs/update_ticket_links_job.rb
LS80: In Ruby a method will return the return value of the last evaluated statement. I don't think it particularly matters here what the method is returning. If you want it to return `nil` here then it would need to be
```suggestion
    unless ticket_key
      post_not_found_status(full_repo_name, after_sha)
      return
    end 
```

spec/jobs/update_ticket_links_job_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+        expect(JiraClient).to receive(:post_comment).with(ticket_key, expected_comment)
+        stub = stub_request(:post, "https://api.github.com/repos/#{full_repo_name}/statuses/#{after_sha}")
+               .to_return(status: 200)
+        subject.perform(args)
+        remove_request_stub(stub)
+      end
+    end
+  end
+
+  describe '#extract_ticket_key_from_branch_name' do
```
LS80: It's not normally a good idea to test private methods as it's too coupled to the implementation which makes it hard to refactor the class.

kylejm: Discussed offline - agreed that regex matching should be tested but not through a private method, rather via it's own class or something similar.

Not going to fix in this PR, however.

app/jobs/update_ticket_links_job.rb
kylejm: Fixed 👍

PR #418, opened 2019-06-17
ENG-2102: Filter 'token' values within logs of HTTP requests

JuubiSnake:
This PR:
* Filters 'token' params from logs and redacts them, labelling the value of each param as [FILTERED]
* Patches rails CommonLogger in order to filter tokens that are included within a HTTP request, since `filter_parameters` does not apply to this class.

This references https://jira.fundingcircle.com/browse/ENG-2102

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'rails', '~> 4.2.11'
-
+gem 'rack', '1.6.11'
 gem 'addressable', require: 'addressable/uri'
 gem 'bootstrap-sass'
 gem 'delayed_job_active_record'
--- a/config/initializers/rack_common_logger.rb
+++ b/config/initializers/rack_common_logger.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+# rubocop:disable all
+
+require 'rack/utils'
+
+# We are going to monkey patch the commonlogger that rack uses in order to
+# filter out any 'token' parameters found within a http request.
+#
+# The common logger can be found here:
+# https://github.com/rack/rack/blob/master/lib/rack/common_logger.rb
+module Rack
+  class CommonLogger
+    # # Common Log Format: http://httpd.apache.org/docs/1.3/logs.html#common
+    # #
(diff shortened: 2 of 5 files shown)


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'rails', '~> 4.2.11'
-
+gem 'rack'
```
fussybeaver: Since we're monkey patching rack, consider pegging the library at whichever version we are monkey patching.

JuubiSnake:
A quick preview with a dummy token when calling the `event` endpoint via curl:

Screenshot 2019-06-21 at 16 14 29

spec/spec_helper.rb
```diff
@@ -48,7 +48,7 @@
     mocks.verify_partial_doubles = true
   end
 
-  config.disable_monkey_patching!
+  # config.disable_monkey_patching!
```
sgerrand: 🤔 Was there a reason to do this?

config/initializers/rack_common_logger.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+# rubocop:disable all
+
+require 'rack/utils'
+
+# We are going to monkey patch the commonlogger that rack uses in order to
```
sgerrand: 🙋 You can swap the logger than Rails/Rack uses at runtime:
> `config.logger` accepts a logger conforming to the interface of `Log4r` or the default Ruby `Logger` class. Defaults to an instance of `ActiveSupport::Logger`.

https://guides.rubyonrails.org/v4.2/configuring.html

You can also [configure which logger is used via Rack middleware](https://guides.rubyonrails.org/v4.2/configuring.html#configuring-middleware) but I wouldn't recommend doing so for logging of this kind.

spec/spec_helper.rb
JuubiSnake: Testing a few things - accidentally pushed with this in, so I'll revert it back

LS80:
We can close this PR. The logging middleware needs to be removed to stop swallowing the Octokit client errors, which means we will no longer be logging the request at all.

PR #423, opened 2019-09-17
Retry commit status updates when GitHub API rate limit is reset

LS80:
When Shipment Tracker makes more than 5000 GitHub API requests per hour, PR status updates will begin to fail and the delayed job queue builds up with retries which will never succeed. Currently there are [10](https://github.com/FundingCircle/shipment_tracker/blob/5ea1759ce8ef8d8647fc434b93b6bf33a71e3855/config/initializers/delayed_job.rb#L4) retries per status update which causes a huge backlog of jobs, until the limit is reset. This creates a lot of log spam. After the retries are exhausted the PR needs to be manually updated to trigger a new status update.

This change retries the job when the GitHub API rate limit is reset, therefore allowing the original job to complete at the earliest opportunity.

#### TODO
- [x] Do a real world test.
- [x] Fix specs and add new examples.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,3 +57,9 @@ COPY --chown=shipment_tracker:shipment_tracker dropzone.yaml /usr/local/deploy/d
 COPY --chown=shipment_tracker:shipment_tracker . .
 
 ENTRYPOINT ["/sbin/tini", "--", "docker-entrypoint.sh"]
+
+ARG REVISION=unknown
+ENV REVISION=$REVISION
+ARG NAME=shipment_tracker
+LABEL name=$NAME version=$REVISION
+RUN echo "$REVISION" > REVISION
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,7 +147,12 @@ jobs:
           command: echo "$DOCKER_PASSWORD" | docker login --username="$DOCKER_USERNAME" --password-stdin $DOCKER_REGISTRY
       - run:
           name: Build Docker image
-          command: docker build . --tag $CIRCLE_PROJECT_REPONAME --pull
+          command: |
+            docker build . \
+              --tag $CIRCLE_PROJECT_REPONAME \
+              --pull \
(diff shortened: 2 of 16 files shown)


Dockerfile
```diff
@@ -57,3 +57,9 @@ COPY --chown=shipment_tracker:shipment_tracker dropzone.yaml /usr/local/deploy/d
 COPY --chown=shipment_tracker:shipment_tracker . .
 
 ENTRYPOINT ["/sbin/tini", "--", "docker-entrypoint.sh"]
+
+ARG REVISION=unknown
+ENV REVISION=$REVISION
```
sgerrand:
```suggestion
```
This line is redundant, the previous line is enough to set the value.

Dockerfile
```diff
@@ -57,3 +57,9 @@ COPY --chown=shipment_tracker:shipment_tracker dropzone.yaml /usr/local/deploy/d
...
+ARG NAME=shipment_tracker
+LABEL name=$NAME version=$REVISION
+RUN echo "$REVISION" > REVISION
```
sgerrand: 💭 I think that labelling the image is better than leaving a file with a commit in the image. What do you think?

Dockerfile
LS80: The idea is to have the `REVISION` environment variable set at container run time, which wouldn't be the case without this line.

Dockerfile
LS80: We need both.

The image `version` label is to override the one from the [base image](https://github.com/FundingCircle/docker-images/blob/3278a63051ce4cb53d0da4dfc1213bde250705a3/alpine/0-envconsul/Dockerfile-3.10#L4).

The `REVISION` file is so that [Loga](https://github.com/FundingCircle/loga/blob/7b18e212daeb75e1205e47d71715a71770b41e2b/lib/loga/service_version_strategies.rb#L7) can set the `service_version` in Graylog, which is currently `unknown.sha`.

PR #424, opened 2019-09-25
Allow disabling github status updates for certain applications

bliof:
This feature is needed as we have a lot of repositories and some
workflows cause a lot of calls to github and we get rate limited.

Work like this:

```
DISABLE_GITHUB_STATUS_UPDATES_FOR_APPS='shipment-tracker, FundingCircle/loga'
```

allowed values: `org/repo` or just the `repo` name

### DIFF
--- a/app/models/commit_status.rb
+++ b/app/models/commit_status.rb
@@ -38,6 +38,8 @@ class CommitStatus
   end
 
   def last_status
+    return unless allow_requests?
+
     @last_status ||= github.last_status_for(repo: full_repo_name, sha: sha)
   end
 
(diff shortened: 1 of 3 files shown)


app/models/commit_status.rb
```diff
@@ -57,6 +61,15 @@ def post_status(notification, target_url = nil)
     )
   end
 
+  def allow_requests?
+    app = full_repo_name.split('/').last
+
+    !(
+      ShipmentTracker::DISABLE_GITHUB_STATUS_UPDATES_FOR_APPS.include?(app) ||
```
sgerrand: 🤔 Do we need to support both forms (org + repo, repo)? It seems like the fully qualified version should be enough.

LS80:
It will help as a short term fix to unblock the teams, but the whole ticket re-linking process needs to be fixed properly.

LS80:
Feels like it would be better to combine it with https://github.com/FundingCircle/shipment_tracker/blob/eff7662395ae4d52f1a5f5120fc670cf6d0dc334/lib/clients/github.rb#L21

Not a blocker as it's (hopefully) a short term fix.

app/models/commit_status.rb
bliof: I've added the `app` as that is the repository name on `/repositories`. It is now loaded from the db as it is too much of a hassle.

sgerrand:
Needs a 🎫 added. 🥁

LS80:
Tested successfully on UAT.

bliof:
✅  Ticket added.

Any code changes or shall we merge it?

@LS80 the config needs refactoring. There are also comments like e.g.:
>  TODO: Move our constants here. Keep Rails config for actual Rails configuration.

sgerrand:
🙏 Please rebase.

PR #425, opened 2019-09-25
Bump ruby to 2.5.6

bliof:
It is already like that in the docker image.


LS80:
The Docker image actually has `2.5.5p157`.

```
docker pull quay.io/fundingcircle/shipment_tracker
docker run --rm  quay.io/fundingcircle/shipment_tracker ruby -v
```

bliof:
replaced by https://github.com/FundingCircle/shipment_tracker/pull/430

PR #427, opened 2019-09-25
Prometheus metrics

sgerrand:
💁 These changes enable a metrics interface using the [`prometheus_exporter`](https://github.com/discourse/prometheus_exporter) gem.

I've enabled metrics for:
* Delayed Job
* process
  * RSS
* Unicorn
  * sessions

🎉 Ready for review. 🙇


dropzone.yaml
```diff
@@ -71,6 +73,33 @@ deploy_tasks:
           volumes:
             - *git_volume
 
+      - id: prometheus-exporter
```
bliof: 🤔 How about having it as part of the `web` instance?

dropzone.yaml
```diff
@@ -71,6 +73,33 @@ deploy_tasks:
...
+            protocol: HTTP
+            gracePeriodSeconds: 30
+            intervalSeconds: 30
+            timeoutSeconds: 20
+            maxConsecutiveFailures: 3
+        labels:
+          tags: http
+        taskKillGracePeriodSeconds: 30
+        constraints:
+          - *az_constraint
```
LS80: Since this process does not use the NFS git volume, we can remove the availability zone constraint.

config/initializers/delayed_job.rb
```diff
@@ -1,5 +1,14 @@
 # frozen_string_literal: true
 
+if Rails.env != 'test'
```
LS80:
```suggestion
unless Rails.env.test?
```

dropzone.yaml
sgerrand: That violates the single process/purpose Docker paradigm, as well as overloading the `web` process. I'd prefer not to do that.

config/initializers/delayed_job.rb
sgerrand: ♻️ Applied in aa5994f3, thanks!

dropzone.yaml
sgerrand: ♻️ Applied in 85d545cf, thanks!

dropzone.yaml
bliof: tl&dr; what we talked in slack is that this process is collecting the metrics pushed by the other processes https://github.com/discourse/prometheus_exporter#multi-process-mode 

and here is the config  https://github.com/FundingCircle/shipment_tracker/blob/85d545cfc4e5aad99469d070d8deef709ade996a/config/initializers/prometheus.rb#L4-L6

dropzone.yaml
```diff
@@ -71,6 +73,31 @@ deploy_tasks:
...
+        container:
+          docker:
+            portMappings:
+              - name: http
+                containerPort: 9394
+                hostPort: 0
+                labels:
+                  metrics: "/metrics"
+        healthChecks:
+          - path: /metrics
```
bliof: 🤷‍♂ Not sure if this is ideal as it will be queried by a lot of tools and not only prometheus. I did a quick review 👀  on the gem and it seems that it just mainly returns collected things so it should be fast. 

https://github.com/discourse/prometheus_exporter/blob/master/lib/prometheus_exporter/server/web_server.rb#L51

docker-compose.yml
```diff
@@ -94,6 +96,19 @@ services:
       web:
         condition: service_healthy
 
+  prometheus-exporter:
+    image: shipment_tracker
+    command: bundle exec prometheus_exporter
```
sgerrand: 👋 @FundingCircle/orchestration-and-routing Should I apply a `--prefix` to these metrics? They'll come through as `ruby_` otherwise.

dropzone.yaml
```diff
@@ -71,6 +73,31 @@ deploy_tasks:
           volumes:
             - *git_volume
 
+      - id: prometheus-exporter
+        instances: 1
+        cpus: 0.25
+        mem: {{ prometheus_exporter_memory }}
+        args: ["bundle", "exec", "prometheus_exporter"]
```
sgerrand: 📝 Also may need a `--prefix` here too.

dropzone.yaml
kamilhristov: +1 perhaps it is not critical for this particular app, but it is better to use just / endpoint if it returns 200 or TCP health check instead.

Some exporters/apps are doing heavy queries on every hit + some metrics like scrape count/errors won't show accurate information since Marathon is also hitting the endpoint.

config/prometheus_client.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+unless Rails.env.test?
+  require 'prometheus_exporter/client'
+  PrometheusExporter::Client.default = PrometheusExporter::Client.new(host: ENV.fetch('PROMETHEUS_EXPORTER_HOST'),
+                                                                      port: ENV.fetch('PROMETHEUS_EXPORTER_PORT'))
+end
```
LS80: Could this configuration be put in `config/initializers/prometheus.rb`? Is there an initialization order issue that made this file necessary?

dropzone.yaml
sgerrand: That makes sense. Thanks for pointing it out. 🙇

docker-compose.yml
bliof: 🤔 for the ruby related metrics it kinda makes sense e.g.:

```
ruby_rss{app="/shipment-tracker/prometheus-exporter", type="web",...
```

For the http ones it is weird

```
ruby_http_duration_seconds{action="index",app="/shipment-tracker/prometheus-exporter",controller="home",image="quay.io/fundingcircle/shipment_tracker:prometheus-metrics_8500",job="marathon",quantile="0.01",ask="shipment-tracker_prometheus-exporter.7678b686-e436-11e9-881d-0242b441c530",version="8c63bfd00d9524d313f71598d57835cb0986b747",...
```

also the ruby_http_duration_seconds doesn't include a `type` for some reason.

Maybe we can call them app_rss and app_http_duration_seconds 🙈 but I don't see a problem with keeping the `ruby_` prefix.

Not sure what we can do about the type, it would be great to have an `app` label that is `app=/shipment-tracker/web`, `app=/shipment-tracker/delayed-job` and so on..

The current app label is the relabel-ing of the __meta_marathon_app -> /shipment-tracker/prometheus-exporter

dropzone.yaml
sgerrand: 🤔 Some further investigation required.
```
$ curl -i 127.0.0.1:9394/
HTTP/1.1 404 Not Found 
Content-Type: text/plain; charset=utf-8
Server: WEBrick/1.4.2 (Ruby/2.5.5/2019-03-15)
Date: Wed, 02 Oct 2019 16:03:14 GMT
Content-Length: 82
Connection: Keep-Alive

Not Found! The Prometheus Ruby Exporter only listens on /metrics and /send-metrics/
```

sgerrand: Using `/send-metrics` should be 👌 though.

sgerrand:
```
$ curl -i 127.0.0.1:9394/send-metrics
HTTP/1.1 200 OK 
Content-Type: text/plain; charset=utf-8
Server: WEBrick/1.4.2 (Ruby/2.5.5/2019-03-15)
Date: Wed, 02 Oct 2019 16:04:41 GMT
Content-Length: 2
Connection: Keep-Alive

OK
```

config/prometheus_client.rb
LS80: Putting this config in `config/environments/production.rb` seems to work. Have a look at https://github.com/FundingCircle/shipment_tracker/commit/bb2901a36a7506be131653deb9afcce7b448ebfd.

sgerrand: > Is there an initialization order issue that made this file necessary?

There is indeed. The `DelayedJob` initialiser gets loaded before this one, which meant that it didn't start with the Prometheus configuration.

LS80: `config/environments/production.rb` is loaded before the initializers though, so it could go there.

LS80:
✅ I've verified that the Delayed Job metrics are collected, by creating a new PR for [hello_world_rails](https://github.com/FundingCircle/hello_world_rails).

```
# HELP ruby_delayed_job_duration_seconds Total time spent in delayed jobs.
# TYPE ruby_delayed_job_duration_seconds counter
ruby_delayed_job_duration_seconds{job_name="AutoLinkTicketJob"} 0.014137454330921173
ruby_delayed_job_duration_seconds{job_name="CommitStatusNotFoundJob"} 0.9399375803768635

# HELP ruby_delayed_jobs_total Total number of delayed jobs executed.
# TYPE ruby_delayed_jobs_total counter
ruby_delayed_jobs_total{job_name="AutoLinkTicketJob"} 1
ruby_delayed_jobs_total{job_name="CommitStatusNotFoundJob"} 1
```

config/initializers/prometheus.rb
```diff
@@ -0,0 +1,16 @@
...
+  require 'prometheus_exporter/middleware'
+
+  # This reports stats per request like HTTP status and timings
+  Rails.application.middleware.unshift PrometheusExporter::Middleware
+
+  require 'prometheus_exporter/instrumentation'
+
+  # this reports basic process stats like RSS and GC info, type master
+  # means it is instrumenting the master process
+  PrometheusExporter::Instrumentation::Process.start(type: 'master')
```
LS80: Shouldn't we have one of these instrumentations each for `web`, `event-worker`, `delayed-job`, and `git-worker`? Then we should have metrics like `ruby_rss` with the correct `type` for each process.

sgerrand: That seems reasonable. I threw the initial exporter/collector in as a centralised point for all the Marathon jobs to report into but perhaps it would be more effective to have separate ones for each job `type`.

sgerrand: It might be difficult to configure that based on the Marathon job. I'll see what values we get by default in the Marathon jobs' environments.

LS80: Yes I'm thinking something like https://github.com/FundingCircle/shipment_tracker/commit/12fc95db8fc7f092ae90991eca39426685d5d533.

Which produces
```
# HELP ruby_rss Total RSS used by process.
# TYPE ruby_rss gauge
ruby_rss{type="git-worker",pid="6"} 110931968
ruby_rss{type="delayed-job",pid="7"} 112439296
ruby_rss{type="event-worker",pid="7"} 112472064
ruby_rss{type="web",pid="20"} 110862336
```

sgerrand: 😻 Nice!

LS80:
@sgerrand I add the "type" label commit. Ready to merge?

Issue #196, opened 2017-01-03
Remove solid_use_case gem


bliof:
Main purpose of the gem is to provide some control flow helpers and that should help in some way the readability of the code. The pattern that it wraps could be accomplished with very basic ruby syntax and I think there is no need it to be enforced with the help of the gem.

The most complex way we use it is: ([adding of repo](https://github.com/FundingCircle/shipment_tracker/blob/master/app/controllers/git_repository_locations_controller.rb#L11-L23), [feature_reviews](https://github.com/FundingCircle/shipment_tracker/blob/master/app/controllers/feature_reviews_controller.rb#L27-L35))

```
AddRepositoryLocation.run(git_repo_location_params).match do
  success do |repo_name|
    ...
  end

  failure do |error|
    ...
  end
end
```

which is an `if` statement. Even if there are more results you could always write it ruby:

```
result = AddRepositoryLocation.call(git_repo_location_params)

case result[:status]
when :success
when :failure
when :almost_success
else
end
```

In the use_case models we only use `continue` and `fail`. e.g. [add repository location](https://github.com/FundingCircle/shipment_tracker/blob/master/app/use_cases/add_repository_location.rb)

bliof:
@0mega 👀  Let's remove the gem?

I'll make a PR where I'll rewrite the AddRepositoryLocation in normal ruby if you think it is ok to remove the gem.

bliof:
ps. my bad we can start discussing on the https://github.com/FundingCircle/shipment_tracker/pull/194 and the other could be updated quickly afterwards

PR #429, opened 2019-09-26
Use pull request events instead of push events

LS80:
Using pull request synchronize events from GitHub instead of push events should fix a few issues caused by unnecessary load from pushes to long lived staging branches. With this change we only take action when a commit is pushed to a PR against `master`. We can then re-enable GitHub status updates for `ce-salesforce` and `salesforce`.

This will require turning on `Pull Requests` events in the [Webhook settings](https://github.com/organizations/FundingCircle/settings/hooks/5440962). We can also turn off `Pushes`.
image


app/controllers/github_notifications_controller.rb
```diff
@@ -1,13 +1,17 @@
 # frozen_string_literal: true
 
-require 'payloads/github'
```
sgerrand: 💭 Did the class name need to change?

LS80: Not really. I just made it more explicit as it represents the payload from that specific event.

sgerrand: Ah, OK.

app/use_cases/handle_pull_request_created_event.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'git_repository_location'
+
+class HandlePullRequestCreatedEvent
+  include SolidUseCase
```
bliof: 🙋‍♂ Let's not use this gem.. The dsl is very strange and we don't really use it. Relates to #196

LS80: Maybe we should remove it, but I won't to do it in this PR. I'm just following the existing patterns.

README.md
```diff
@@ -93,12 +93,19 @@ In addition to the `update_git_loop` task, you can set the environment variable
...
+a Shipment Tracker status check.
+
+To configure the webhook
+ 1. In GitHub, add a new webhook in the repository or organisation settings.
+ 2. In the API Tokens tab of Shipment Tracker, find (or create) a Github Notifications token.
+ 3. Set "Payload URL" to the GitHub notifications URL from Shipment Tracker.
+ 4. Set "Content type" to "application/json".
+ 5. Select "Let me select individual events."
+ 6. Select "Pull requests".
+ 7. Click "Add webhook"
```
sgerrand: 👏 Thank you for the documentation 💅 .

app/jobs/relink_ticket_job.rb
```diff
@@ -11,10 +11,8 @@ def perform(args)
     before_sha = args.delete(:before_sha)
     after_sha = args.delete(:after_sha)
     full_repo_name = args.delete(:full_repo_name)
-    branch_created = args.delete(:branch_created)
 
-    if branch_created || commit_on_master?(full_repo_name, after_sha) ||
-       relink_tickets(before_sha, after_sha).empty?
+    if relink_tickets(before_sha, after_sha).empty?
```
sgerrand: 👏

app/models/payloads/github_pull_request.rb
```diff
@@ -0,0 +1,65 @@
...
+    def full_repo_name
+      pull_request&.dig('base', 'repo', 'full_name')
+    end
+
+    def branch_name
+      pull_request&.dig('head', 'ref')
+    end
+
+    def base_branch_master?
+      pull_request&.dig('base', 'ref') == 'master'
```
sgerrand: 👏

PR #435, opened 2019-10-14
Add Unicorn Prometheus metrics

LS80:
```
ruby_unicorn_active_workers_total
ruby_unicorn_request_backlog_total
ruby_unicorn_workers_total
```


config/unicorn.rb
```diff
@@ -1,9 +1,23 @@
...
+require 'prometheus_exporter/instrumentation'
+require_relative 'prometheus_client'
+
+port_http = ENV.fetch('PORT_HTTP')
+pid_file = '/tmp/unicorn.pid'
+
+PrometheusExporter::Instrumentation::Unicorn.start(
+  pid_file: pid_file,
+  listener_address: "0.0.0.0:#{port_http}",
+  client: nil # Use default client
```
sgerrand:
```suggestion
```
The implementation in https://github.com/FundingCircle/shipment_tracker/blob/2e48e4e90d18fdb86f2727aae0e27b7d06966f62/config/initializers/delayed_job.rb#L5 didn't pass a `nil` `client:` in. Should that aspect of the configuration be the same?

LS80: Yes it shouldn't be made a required keyword argument if it actually has a default
https://github.com/discourse/prometheus_exporter/blob/fc8758cb6c749e281d893620483306477a7e4e70/lib/prometheus_exporter/instrumentation/unicorn.rb#L12.

LS80:
Yes indeed that's why I haven't merged it yet, as it's misleading. I'll probably have to change the Unicorn instrumentation to be able to add a unique label per web process, so that the values can be aggregated in Prometheus. I'll try to make a change upstream instead of monkey patching it.

bliof:
The code of the gem mentions `hostname` and some have `pid` but most probably the easiest is to specify both manually https://github.com/discourse/prometheus_exporter#metrics-default-prefix--labels

LS80:
That won't work because we aren't using the single process mode https://github.com/discourse/prometheus_exporter/pull/20.

I've opened https://github.com/discourse/prometheus_exporter/pull/90.

config/unicorn.rb
LS80: I've opened https://github.com/discourse/prometheus_exporter/pull/90.

LS80: discourse/prometheus_exporter#90 has been merged. I've bumped the gem to use the master branch until there's a new release.

We now have a hostname label for the metrics.
```
# HELP ruby_unicorn_workers_total Number of unicorn workers.
# TYPE ruby_unicorn_workers_total gauge
ruby_unicorn_workers_total{hostname="e2cf40e4314e"} 4
ruby_unicorn_workers_total{hostname="ca61920edc58"} 4
```

LS80:
discourse/prometheus_exporter#90 has been merged. I've bumped the gem to use the master branch until there's a new release.

We now have a hostname label for the metrics.
```
# HELP ruby_unicorn_workers_total Number of unicorn workers.
# TYPE ruby_unicorn_workers_total gauge
ruby_unicorn_workers_total{hostname="e2cf40e4314e"} 4
ruby_unicorn_workers_total{hostname="ca61920edc58"} 4
```

@bliof

LS80:
The `git-worker` is crashing so this is blocked for now. I only see a huge core dump in Graylog so no idea what the problem is yet.

LS80:
Problem fixed by bumping the library versions in the Docker image.

PR #439, opened 2019-11-07
Switch deploy_alerts to text

bliof:
They are a text field and can contain very large messages for new apps


LS80:
Strange because `character varying` and `text` should both accept unlimited length strings.

LS80:
From https://www.postgresql.org/docs/9.6/datatype-character.html

> If character varying is used without length specifier, the type accepts strings of any size

> If you desire to store long strings with no specific upper limit, use text or character varying without a length specifier, rather than making up an arbitrary length limit.

bliof:
No idea.. the error is:

```
ActiveRecord::StatementInvalid: PG::ProgramLimitExceeded: ERROR:  index row requires 45608 bytes, maximum size is 8191
: UPDATE "deploys" SET "deploy_alert" = $1 WHERE "deploys"."id" = $2
```

alexdias:
@bliof https://stackoverflow.com/questions/46139581/postgresql-btree-index-row-size-limitation

Basically, you can't index something that big.

LS80:
Yeah it's the index that's the problem

bliof:
Let's drop the index 🙈

bliof:
Follow up PR https://github.com/FundingCircle/shipment_tracker/pull/440

PR #440, opened 2019-11-07
Drop the deploy_alert index

bliof:
When trying to store long deploy alerts the index crashes with a size limit

```
ActiveRecord::StatementInvalid: PG::ProgramLimitExceeded: ERROR:  index row requires 45608 bytes, maximum size is 8191
: UPDATE "deploys" SET "deploy_alert" = $1 WHERE "deploys"."id" = $2
```

### DIFF
--- a/db/migrate/20191107163219_drop_deploys_deploy_alert_index.rb
+++ b/db/migrate/20191107163219_drop_deploys_deploy_alert_index.rb
@@ -1,0 +1,9 @@
+class DropDeploysDeployAlertIndex < ActiveRecord::Migration
+  def up
+    remove_index :deploys, :deploy_alert
+  end
+
+  def down
+    add_index :deploys, :deploy_alert
+  end
+end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2,26 +2,19 @@
 -- PostgreSQL database dump
 --
 
+-- Dumped from database version 9.6.15
+-- Dumped by pg_dump version 11.5
+
 SET statement_timeout = 0;
 SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
 SET client_encoding = 'UTF8';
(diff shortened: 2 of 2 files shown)


bliof:
It is used once in `Repositories::DeployRepository#unapproved_production_deploys_for`

db/structure.sql
```diff
@@ -2,26 +2,19 @@
 -- PostgreSQL database dump
 --
 
+-- Dumped from database version 9.6.15
+-- Dumped by pg_dump version 11.5
```
LS80: Is this `structure.sql` backward compatible with `9.6`? Probably best to generate it with the same `pg_dump` version.

db/migrate/20191107163219_drop_deploys_deploy_alert_index.rb
```diff
@@ -0,0 +1,9 @@
+class DropDeploysDeployAlertIndex < ActiveRecord::Migration
+  def up
+    remove_index :deploys, :deploy_alert
+  end
+
+  def down
+    add_index :deploys, :deploy_alert
+  end
+end
```
LS80: Just noticed that in Rails 5 `remove_index` is reversible so you can use `change` instead of `up` and `down`. Not so for Rails 4.2.

db/structure.sql
bliof: I won't be able to do that today, if you can pick it up it would be great.

bliof: Otherwise I think pg_dump/pg_restore/psql are fairly backward compatible. TBO not sure what will happen if you try to restore with an older version but using the newer should work.

bliof:
💭 Maybe we can use a GIST index but that is for another PR.

db/structure.sql
LS80: I'll check...

PR #444, opened 2020-01-28
Small improvements

bliof:
*tl; dr*

- Refactored development flow so everything kinda works
  - ✅  `docker-compose up`
  - ✅  `docker-compose up jira postgres;` + `foreman start`
  - ✅  `docker-compose up jira postgres` + `./bin/rails server`
- Refactored the use of the prometheus gem.
- "Fix" reaping bug
- There is also a `SKIP_AUTHENTICATION` env that when added will log you in with `admin@example.com`

# Fixes:

- [x] bump the ruby in .ruby-version to 2.5.7
- [x] fix the `docker-compose up` so it doesn't complain about SECRET_KEY_BASE. It is loaded from the secrets.env
- [x] `docker-compose up` crashes for 3-4 times and then starts working - the healthchecks were telling that one of the web containers is unhealthy and that caused the flakiness 
```
$ docker-compose up
Starting shipment_tracker_postgres_1 ... done
Starting shipment_tracker_jira_1     ... done
Starting shipment_tracker_prometheus-exporter_1 ... done
Starting shipment_tracker_delayed-job_1         ...
Starting shipment_tracker_delayed-job_1         ... done
Starting shipment_tracker_web_1                 ... done
Starting shipment_tracker_web_2                 ... done
Starting shipment_tracker_event-worker_1        ... done

ERROR: for git-worker  Container "59ec6b4627fe" is unhealthy.
ERROR: Encountered errors while bringing up the project.
```
  * don't use scale and have 2 separate web nodes
- [x] doing foreman start or ./bin/rails s doesn't work ?? when you open the localhost:5400 or localhost:3000 nothing shows..
  * just using `thin` does the trick
- [x] double check whether the prometheus exporter works on mac os - seems to work
```
Prometheus Exporter Failed To Collect Unicorn Stats undefined method `tcp_listener_stats' for Raindrops::Linux:Module
```
 * works well, the PR includes refactoring and extracting all calls to the gem in a separate module
- [x] reduce the amount of `reap`-ing by unicorn
 * remove the unicorn instrumentation as it is buggy - the one that shows number of workers


docker-compose.yml
```diff
@@ -28,7 +28,6 @@ services:
       JIRA_USER: shipment_tracker
       JIRA_USE_SSL: 'false'
       PORT_HTTP: 1201
-      SECRET_KEY_BASE: "$SECRET_KEY_BASE"
```
LS80: It's needed for https://github.com/FundingCircle/shipment_tracker/blob/52bf6e07521824e0cfb8ea3445d2239ca7406265/config/secrets.yml#L21-L22 but I suppose you are setting it in `secrets.env`?

.ruby-version
```diff
@@ -1 +1 @@
-2.5.5
+2.5.7
```
LS80: There is a Ruby 2.6 image now if you want to try that.

docker-compose.yml
LS80: Missed your explanation in the description, ignore the above 😄

.ruby-version
bliof: Maybe in another PR 🙈

docker-compose.yml
```diff
@@ -46,8 +46,11 @@ services:
     volumes:
       - git:/tmp/git
     healthcheck:
-      test: curl localhost:1201/healthcheck
+      test: curl -fsSL "http://localhost:1201/healthcheck" || exit 1
```
LS80: It would be helpful to use the long form options here, especially `--fail` and `--show-error`.

app/controllers/concerns/authentication.rb
```diff
@@ -21,8 +21,12 @@ def unauthenticated_strategy
...
 
   def current_user
-    setup_current_user!
-    User.new(session[:current_user])
+    if ENV['SKIP_AUTHENTICATION'] == 'true'
+      User.new(first_name: 'Admin', email: 'admin@example.com')
+    else
+      setup_current_user!
+      User.new(session[:current_user])
+    end
```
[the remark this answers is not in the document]
bliof: When you run with `foreman start` / `bin/rails s` it is already in `if Rails.env.development?` but that doesn't help. Do you suggest it just for symmetry -> to make sure the that the oauth is disabled?

LS80:
The `reap`ing issue could be caused by https://github.com/FundingCircle/shipment_tracker/blob/52bf6e07521824e0cfb8ea3445d2239ca7406265/config/unicorn.rb#L13

It's much more frequent than for the example usage.

bliof:
@LS80 What I don't get is way it says `worker=unknown` 🤔 

```
web_1                  | I, [2020-01-29T16:05:36.064811 #6]  INFO -- : reaped #<Process::Status: pid 10259 exit 0> worker=unknown
web_2                  | I, [2020-01-29T16:05:36.208743 #6]  INFO -- : reaped #<Process::Status: pid 10227 exit 0> worker=unknown
```

also the `docker-compose up` still crashes from time to time and when it crashes it sometimes starts 3 web containers 🤪

bliof:
So.. the unicorn instrumentation somehow bugs the unicorn master process 

```
 8966 root      0:10 {bundle} unicorn master -c config/unicorn.rb
 9241 root      0:00 {bundle} unicorn worker[0] -c config/unicorn.rb
 9245 root      0:00 {bundle} unicorn worker[1] -c config/unicorn.rb
 9253 root      0:00 {bundle} unicorn worker[2] -c config/unicorn.rb
 9259 root      0:00 {bundle} unicorn worker[3] -c config/unicorn.rb
```

but constantly there is a rouge process pop-ing up

```
10753 root      0:00 [unicorn.rb:15]
```

I've tried:

```
  # reaps all unreaped workers
  def reap_all_workers
    puts "reap all workers!!!!!!!!!!!!!!!1"
    begin
      wpid, status = Process.waitpid2(-1, Process::WNOHANG)
      wpid or return
      if @reexec_pid == wpid
        logger.error "reaped #{status.inspect} exec()-ed"
        @reexec_pid = 0
        self.pid = pid.chomp('.oldbin') if pid
        proc_name 'master'
      else
        puts "wpid: #{wpid}"
        worker = @workers.delete(wpid) and worker.close rescue nil
        @after_worker_exit.call(self, worker, status)
      end
    rescue Errno::ECHILD
      puts "echild"
      break
    end while true
  end
```

and it constantly spams "reap all workers!!!!!!!!!!!!!!!1" tbo still no idea why N.B it is before the begin block. Also it never enters the echild clause.

https://github.com/defunkt/unicorn/blob/d03dd4e9e4ff29689752b7c82202008fefaf1210/lib/unicorn/http_server.rb#L397

bliof:
@LS80  How about merging this PR? 

For the docker-compose up bug, I've removed the `scale` and now there is web and web-02 and docker-compose starts them one by one and it seems to work.

For the `reap` but I've removed the unicorn instrumentation.

bliof:
ps. in a follow up PR I'll try to improve more / rework a bit the prometheus metrics and e.g. add tests and so on.

PR #445, opened 2020-02-04
PagerDuty integration

robertstettner:
This integration is now using the `v0.0.7` version of the Terraform Module, which outputs the PagerDuty integrations URL instead of the token itself.
```
Outputs:

pagerduty_service_integrations = https://funding-circle-usa.pagerduty.com/service-directory/PS1URJ6/integrations
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,7 +167,7 @@ jobs:
               quay.io/fundingcircle/docker-scripts:latest docker/ci-docker-publish
   pagerduty_plan:
     docker:
-      - image: quay.io/fundingcircle/circleci-terraform:latest
+      - image: quay.io/fundingcircle/circleci-terraform:5.3-0.11.14
         auth:
           username: $DOCKER_USERNAME
           password: $DOCKER_PASSWORD
@@ -185,7 +185,7 @@ jobs:
 
   pagerduty_apply:
     docker:
-      - image: quay.io/fundingcircle/circleci-terraform:latest
+      - image: quay.io/fundingcircle/circleci-terraform:5.3-0.11.14
         auth:
           username: $DOCKER_USERNAME
           password: $DOCKER_PASSWORD
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -175,7 +175,7 @@ jobs:
           PLUGIN_ACTIONS: plan
           PLUGIN_ROOT_DIR: terraform/
           TF_VAR_service_name: shipment_tracker
-          TF_VAR_escalation_policy_name: Global DevOps Escalation Policy
+          TF_VAR_escalation_policy_name: Observability
 
     steps:
       - checkout
@@ -192,7 +192,7 @@ jobs:
         environment:
           PLUGIN_ROOT_DIR: terraform/
           TF_VAR_service_name: shipment_tracker
-          TF_VAR_escalation_policy_name: Global DevOps Escalation Policy
+          TF_VAR_escalation_policy_name: Observability
 
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -220,13 +220,13 @@ workflows:
             - features
       - publish-docker-image:
           context: org-global
-      - pagerduty_plan:
-          context: org-pagerduty
-          filters:
-            branches:
-              ignore: master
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -165,6 +165,40 @@ jobs:
               --env DOCKER_PASS="$DOCKER_PASSWORD" \
               --volume /var/run/docker.sock:/var/run/docker.sock \
               quay.io/fundingcircle/docker-scripts:latest docker/ci-docker-publish
+  pagerduty_plan:
+    docker:
+      - image: quay.io/fundingcircle/circleci-terraform:latest
```
sgerrand: 🙋 The source for this image needs to be in version control and the image tagged with the Terraform version it contains.

.circleci/config.yml
```diff
@@ -165,6 +165,40 @@ jobs:
...
+        auth:
+          username: $DOCKER_USERNAME
+          password: $DOCKER_PASSWORD
+        environment:
+          PLUGIN_ACTIONS: plan
+          PLUGIN_ROOT_DIR: terraform/
+          TF_VAR_service_name: shipment_tracker
+          TF_VAR_escalation_policy_name: Global DevOps Escalation Policy
```
sgerrand: 🤔 Suggest creating a policy specific to this application.

.circleci/config.yml
```diff
@@ -165,6 +165,40 @@ jobs:
...
+        environment:
+          PLUGIN_ROOT_DIR: terraform/
+          TF_VAR_service_name: shipment_tracker
+          TF_VAR_escalation_policy_name: Global DevOps Escalation Policy
+
+    steps:
+      - checkout
+      - run:
+          name: Terraform Apply
+          command: /bin/circleci-terraform
```
sgerrand: 🤔 Should this command differ from the previous "Terraform Plan" job? It's not clear what the executable is doing.

.circleci/config.yml
```diff
@@ -186,3 +220,13 @@ workflows:
             - features
       - publish-docker-image:
           context: org-global
+      - pagerduty_plan:
+          context: org-pagerduty
+          filters:
+            branches:
+              ignore: master
+      - pagerduty_apply:
```
sgerrand: 🙋 Perhaps execution of this job should be dependant on all the test jobs passing. What do you think?

.circleci/config.yml
robertstettner: We should have escalation policies resembling org units/owners (for accountability) instead of services. As you will find that you will be changing the escalation policy all the time when there is a re-org. A handover of a service also could add to more disruption with escalation policies per service. It would be simpler to change the escalation policy in that circumstance.

.circleci/config.yml
robertstettner: One entrypoint, all that is changed is the environment variables.

.circleci/config.yml
robertstettner: I agree. :)

.circleci/config.yml
robertstettner: Yes. I have tracked this here: https://jira.fundingcircle.com/browse/PLATFORM-914

.circleci/config.yml
robertstettner: @sgerrand who would effectively own `shipment_tracker` moving forward? Would it be the CD & Tooling team? or the whole of Platform Domain?

.circleci/config.yml
robertstettner: All the relevant documentation and source code for the Docker image is available [here](https://github.com/FundingCircle/circleci-terraform).

.circleci/config.yml
sgerrand: Makes sense. Let's discuss it further out of band.

terraform/backend.tf
```diff
@@ -0,0 +1,11 @@
+terraform {
+  required_version = "= 0.11.14"
```
bliof: 🙋‍♂ why don't you use the 0.12.16 version here and in the image?

robertstettner: Done.

robertstettner:
Bump

sgerrand:
♻️ Rebased.

PR #446, opened 2020-02-13
Remove better_errors Gem

gabrielfunding:
While debugging a production issue with @bliof, we noticed that the error was impossible to reproduce in our local environment. @bliof discovered that `better_errors` was basically preventing us from replicate the issue that we were seeing in production.

This PR will remove this Gem in order to have an environment that replicates production as much as possible.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,8 +38,6 @@ group :dockerignore do
 end
 
 group :development do
-  gem 'better_errors'
-  gem 'binding_of_caller'
   gem 'foreman', require: false
   gem 'spring'
   gem 'spring-commands-rspec'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -44,12 +44,6 @@ GEM
       descendants_tracker (~> 0.0.4)
       ice_nine (~> 0.11.0)
       thread_safe (~> 0.3, >= 0.3.1)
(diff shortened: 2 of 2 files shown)


sgerrand:
> that `better_errors` was basically preventing us from replicate the issue that we were seeing in production

@gabrielfunding, would you or @bliof please provide more details on how `better_errors` made debugging (or any other area) more difficult. I'm not opposed to removing unnecessary libraries but it would be useful for others to understand how this gem was problematic for you both.

bliof:
@sgerrand I've been helping @gabrielfunding fix the double render bug (redirect_to + render) in https://github.com/FundingCircle/shipment_tracker/blob/master/app/controllers/releases_controller.rb#L12-L19

✅  Without better_errors when you open the page locally you get the same error as in production + some stack traces. 

Screenshot 2020-02-13 at 17 04 56

🔴 With better_errors it crashes in the inner error and never hits the rescue block -> in the browser you see an error for https://github.com/FundingCircle/shipment_tracker/blob/b509d1e84e57ed60879b40d5e57d25cf0539be38/app/models/git_repository_loader.rb#L49

Screenshot 2020-02-13 at 17 05 47

sgerrand:
Thank you for that additional context. It's very useful! 🙇

PR #498, opened 2021-06-21
[Bug fix] Main branch issue

mugx-fc:
This PR tries to fix the issue that comes when the main branch is called `main` and not `master`.

### DIFF
--- a/app/models/payloads/github_pull_request.rb
+++ b/app/models/payloads/github_pull_request.rb
@@ -59,7 +59,7 @@ module Payloads
     end
 
     def base_branch_master?
-      pull_request&.dig('base', 'ref') == 'master'
+      %w[main master].include? pull_request&.dig('base', 'ref')
     end
 
     def title


fc-dropzone-github:
### :lock: Prisma Cloud scan details - vulnerabilities :warning:
---
<details open>
<summary><b>:white_check_mark: Passed <code>shipment_tracker:latest</code></b><br />&emsp;&emsp;&ensp;Vulnerabilities found. Total: 0, critical: 0, high: 0, medium: 0, low: 0.</summary><br />




#### No vulnerabilities found :tada:

</details>

app/models/payloads/github_pull_request.rb
```diff
@@ -59,7 +59,7 @@ def branch_name
     end
 
     def base_branch_master?
-      pull_request&.dig('base', 'ref') == 'master'
+      pull_request&.dig('base', 'ref') == 'master' || pull_request&.dig('base', 'ref') == 'main'
```
LS80, line 62:
```suggestion
      %w(main master).include? pull_request&.dig('base', 'ref')
```
That's probably more idiomatic Ruby?

mugx-fc: yep, even better, thanks 👍

LS80 (approved):

0mega (commented):
Hey @mugx-fc, nice to see some love for this repo.

I can see that this is the second attempt to bring `main` branch support and it would be great to get it right. These things are on my mind:

- There are several levels of tests which specify the behaviour of this project regarding branching. We should have an update to those to reflect new functionality. That's the easiest way one has to actually prove that the fix works.
- There are more than a 100 references to `master` in the codebase which would be nice to rename, not a must functionality-wise but if we are serious about fixing this then we should at least have a follow-up PR/Issue for renaming it all.
- There are rake tasks that produce CSV reports based on `master` branch which should also be updated
- Supporting `main` branch actually smells like a subset of the larger problem of supporting custom default branches on GitHub, so instead of hardcoding another value, a different approach could be to provide an array of branches as configuration to this project and provide a friendlier usability.

Let me know your thoughts @mugx-fc, @LS80, @bliof.

mugx-fc:
@0mega agreed on all your points 👍

